PR #1, opened 2018-10-27
Add German translation

michaelherger:


CDrummond:
Thanks for this. Not really sure its ready for translation - as I'm having "fun" using SlimBrowese - and this will change a few things (possibly remove some strings, etc.)

PR #3, opened 2018-10-28
Added Dutch translation

bakker:


CDrummond:
Thanks. But please be aware there may be many changes to come, as the skin is still in the very early stages.

bakker:
No worries, I'll keep my fork in sync and I'll try to keep up with the translations :-)

PR #9, opened 2018-11-13
Desktop pagehandler

michaelherger:
Maybe there's more code to tweak for the different paths in mobile vs. desktop. But this got it more or less working.

I somehow can't get git to recognize the file move. It considers it an addition/deletion, mucking with `desktop.html`'s history. But oh well... the only changes in there are only the changed paths.

BTW: there are two `blank.css` loaded. Why anyway?


CDrummond:
Thanks.  I'll take your suggestion and add a mobile.html, then have index.html (+ some javascript) redirect to desktop/mobile.
---

As to the 2 blank.css. They are dummy placeholders. One gets replaced by the scrollbar css file, and one gets replaced by the dark/light css file. Can't jus tuse one file, as I don't want the scrollbar mods on mobile/touch devices (it forces a scrollbar to be shown there). There's probably a better way to do this, but I'm no web develper - so this is the easiest way I could see.

PR #10, opened 2018-11-13
Add cache-buster in page handler.

michaelherger:
By using the plugin version rather than the individual library versions you wouldn't have to maintain those references. This wouldn't save all the pain for your, but for most users it should work just fine. The cached copies would be refreshed whenever you release a new version.

You could then add a similar handler for the mobile version.

**Please note** that this change **is not complete** at all. You'd have to replace all library specific numbers with the template snippet, plus add it to all of your own JS file references. But I guess you get the idea.

The Perl part would get your plugin's version number and pass it to the page renderer as a new variable `material_revision`. The `[% material_revision %]` then does put that string in place in the page template. (we're using the Perl [Template Toolkit](http://www.template-toolkit.org/docs/manual/Directives.html#section_GET) if you're interested in the details about the syntax)

### DIFF
--- a/MaterialSkin/HTML/material/desktop.html
+++ b/MaterialSkin/HTML/material/desktop.html
@@ -13,7 +13,7 @@
     <title>Logitech Media Server</title>
     <link href="html/css/blank.css" rel="stylesheet">
     <link href="html/css/blank.css" rel="stylesheet">
-    <link href="html/font/font.css" rel="stylesheet">
+    <link href="html/font/font.css?r=[% material_revision %]" rel="stylesheet">
     <link href="html/lib/vuetify.min.css?ver=1.3.5" rel="stylesheet">
     <!-- TODO: Fix and re-use virtual scroller -->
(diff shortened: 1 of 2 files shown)


CDrummond:
The release script (`mkrel.py`) already adds `?ver=<plugin ver>` to the skin's own JS and CSS files - so released versions of this skin should have this already. Not sure the font should be versioned, as its not going to change much. Likewise with the library JS files, these only need updating when their version changes - not when the plugin changes. This is why they have a version specified, and the skin's don't (as script will add).

However, I will (later) merge this - as its more in the LMS way of doing things. Will probably only add version to skin files, not library files.

Not merging now - as I will need to also update the code the changes 'blank.css' so that it contains the version query... (And merging later will help me remember that!)

michaelherger:
Ah, didn't know about your script. And as I said: this is not a pull worthy PR. More of inspiration :-)

CDrummond:
Well merged anyway! Will make changes now...

PR #11, opened 2018-11-15
Check for the existence of MusicArtistInfo without causing an error b…

michaelherger:
…eing logged on the server side.


CDrummond:
Thanks! I did think he existing check was not quite correct...

PR #15, opened 2018-11-20
Use "delete" icon rather than "clear_all". I never understood the lat…

michaelherger:
…ter...


CDrummond:
I can see why the clear icon is not obvious. It is to me, as the music app I use on android uses the same icon. Nut, not sure about "delete" - as it does not delete the playlist. At some point (if I can find space in the UI!) I'd like to show the currently loaded playlist name. Then this would make it look like the lists gets delete, and it does not.

Perhaps just "clear" ??

michaelherger:
Yes, `clear` I'd still prefer over `clear_all` :-). And yes, quite obviously I'm not an Android user...

CDrummond:
Changed icon to "clear"

PR #17, opened 2018-11-20
Move Manage Players item to players menu

michaelherger:
Might be personal preference. Feel free to dismiss. But I believe that the `Manage Players` item should be in the players menu, rather than settings. I actually never saw it there, until I decided to click any item I could find. But I thought it would be nice to have quick access to various aspects of players right from the player menu.

Oh, and this is my first Vue experience. Yay! Feel free to fix it :-D

### DIFF
--- a/MaterialSkin/HTML/material/html/js/toolbar.js
+++ b/MaterialSkin/HTML/material/html/js/toolbar.js
@@ -34,6 +34,11 @@ Vue.component('lms-toolbar', {
     </v-list-tile>
    </template>
 
+   <v-divider v-if="players && players.length>1"></v-divider>
+   <v-list-tile v-if="players && players.length>1" @click="menuAction(TB_MANAGE_PLAYERS.id)">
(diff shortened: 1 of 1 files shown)


CDrummond:
This one I'm not sure about. If this is moved there, perhaps so should player settings? But then that's a setting... Also, if the user has lots of players, it will then be at the bottom of the menu.

michaelherger:
I think settings and player control are different features. One is everyday use, the other only every now and then. I'd therefore still prefer to have the manager with the players rather than in a settings menu. This way I have one menu to go to if I want to control playback of any of my players, not two. But as I said: this is just my personal preference.

CDrummond:
I've merged this now. Now that this is there, do you think I should remove the "Switch On/Off" menu item? As in this functionality can be achieved via the dialog. Only asking as I never actually switch my players off - seeing as they are not real squeezeboxes.

michaelherger:
Thanks! No, I think the power button must stay, as the manager would only be shown if there were more than one player. Plus it's to control the current player, like play/pause/volume.

PR #19, opened 2018-11-23
Add support for virtual and remote library menus

michaelherger:
The menu items to select a library view or browse remote sources (if that plugin is enabled) are missing.

That said: library support seems to be broken currently. Somehow the `library_id` is never sent to the server. I'd always get the full library.

### DIFF
--- a/MaterialSkin/HTML/material/html/js/browse-page.js
+++ b/MaterialSkin/HTML/material/html/js/browse-page.js
@@ -1199,6 +1199,18 @@ var lmsBrowse = Vue.component("lms-browse", {
                                                       icon: "shuffle",
                                                       id: TOP_RANDOM_MIX_ID,
                                                       weight: c.weight ? parseFloat(c.weight) : 100 });
+                            } else if (c.id == "opmlselectVirtualLibrary" || c.id == "opmlselectRemoteLibrary") {
+                                var command = this.buildCommand(c, "go", false, true);
+                                this.serverTop.push({
+                                    title: c.text,
+                                    command: command.command,
+                                    params: command.params,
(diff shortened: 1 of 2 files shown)


CDrummond:
Before accepting this, how do remote libraries work? As in if I navigate into this item, I assume I then get a list of remote libraries. But do I then navigate from there into the library listing? e.g.

My Music
- Artists
- - Albums
...
- Remote Libraries
- - Some Server
- - - Artists
- - - -Albums
..etc ?

If so, then its OK. If (as happens with the web UI and virtual libraries), it changes how the "My Music/Artists" (etc.) lists work (i.e. they now work on the remote library, not local), then I'm against it. In this case the remote library selection needs to be in a settings dialog.

Also, even if it is like the above, I think it belongs in the "Other Sources" section, not under "My Music"

---

I have a fix for for virtual libraries, which I'll commit later.

michaelherger:
Yes, those remote libraries are basically just a navigation tree, like from the music services. Remote libraries could be a UPnP source, or another LMS instance.

As for the library view: I didn't see it in the settings. Is this part of the brokenness? I did see code there, but never got to see the libraries themselves.

CDrummond:
OK. So "Remote Libraries" belongs under "Other Music" Seeing as this will be a hard-coded item (like Favorites, etc) it can't come from that listing. I'll update the "Other Music" section to add this.

Library selection appears in the "Settings" dialog, but only if there is a virtual library. This is working for me. ui-settnigs.js calls "libraries" to obtain the list of library IDs and names. If one or more are returned, then the settings dialog adds a "Default" library selection too.

CDrummond:
I've (hopefully) fixed virtual libraries, and added support for remote libraries.

PR #44, opened 2018-12-20
Only enable artist picture mode if artist pictures are enabled in the…

michaelherger:
… MusicArtistInfo prefs and the MAI plugin is actually running. If we don't respect the plugin availability, we might end up with a grid of placeholder images, if the user ever had the plugin installed before. Prefs are not uninstalled when plugins are.


CDrummond:
Thanks!

PR #46, opened 2018-12-22
Do NOT hide player controls when a player is powered off.

michaelherger:
I know that you expressed your intention not to add this. But really: in mobile mode it now takes at least three taps to start playback on a powered off Radio. With the controls there, it's one. The minimum there can be. Please :-).

I'd have to 
* tap player selector
* select player 
* tap player selector again
* power on
* start playback


michaelherger:
Thank you very much!

PR #71, opened 2019-02-01
Overlay whilst scrolling

sanc909:
Hi

For your consideration: this should be all of the code to create the overlay whilst scrolling .  Never done this before so let me know if it doesn't work. 

Some points to note: 
1. At the moment it  tries to work on all scrollable lists in the skin, so probably needs a bit more intelligence. 
2. Stopwords are hardcoded (removing The/EL) from titles. Could be a parameter for the user to fill in
3. Scrolling long lists in Grid and List view pauses whilst it loads the next batch of entries - didn't want to change your code. 

Requires 
1. A div that can be displayed or not displayed (browse-page.js)
2. Some styling to try and get it to center on the available screen space (desktop.css)
3. Additional code in the listener module (browse-page-.js) to work out what to display. That is
3.1 How many rows have been scrolled,
3.2  Work out which item is at the top of the screen
3.3. Get first letter and display. 

I'm not a vuejs expert so it might be possible to do this differently. 

In grid view it could  show the album cover...

Thanks
San

### DIFF
--- a/MaterialSkin/HTML/material/html/css/desktop.css
+++ b/MaterialSkin/HTML/material/html/css/desktop.css
@@ -11,6 +11,16 @@
     --bottom-toolbar-height: 80px;
 }
 
+
+#letterOverlay{                               /* Overay displays when scrolling lists */
+    z-index:2;
+    position: fixed;
+    top: calc(var(--main-toolbar-height) + ((((var(--vh, 1vh)*100) - (var(--main-toolbar-height) + var(--bottom-toolbar-height)))/2) - 36px));
+    left: calc(((100vw * (var(--splitter-pc)/100))/2) - 36px);
+}
+
+
+
 body {
     min-width:600px;
     min-height:400px;
--- a/MaterialSkin/HTML/material/html/js/browse-page.js
+++ b/MaterialSkin/HTML/material/html/js/browse-page.js
@@ -141,6 +141,7 @@ var lmsBrowse = Vue.component("lms-browse", {
   </v-layout>
  </div>
  <v-progress-circular class="browse-progress" v-if="fetchingItems" color="primary" size=72 width=6 indeterminate></v-progress-circular>
+ <div v-show="showLetter" id="letterOverlay"></div>
 
  <v-list v-if="useGrid" class="lms-image-grid noselect bgnd-cover" id="browse-grid">
   <div v-for="(item, index) in items" :key="item.id" :id="'item'+index">
@@ -295,7 +296,8 @@ var lmsBrowse = Vue.component("lms-browse", {
             libraryName: undefined,
             selection: [],
             collapsed: [false, false, false]
-        }
+            showLetter: false,                             //toggles display of overlay when scrolling  
+    }
     },
     computed: {
         darkUi () {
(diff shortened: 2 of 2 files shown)


MaterialSkin/HTML/material/html/css/desktop.css
```diff
@@ -11,6 +11,16 @@
     --bottom-toolbar-height: 80px;
 }
 
+
```
CDrummond: Should there not be a similar entry for mobile.css?

MaterialSkin/HTML/material/html/js/browse-page.js
```diff
@@ -1595,17 +1597,53 @@ var lmsBrowse = Vue.component("lms-browse", {
...
+
                 const scrollY = this.scrollElement.scrollTop;
                 const visible = this.scrollElement.clientHeight;
                 const pageHeight = this.scrollElement.scrollHeight;
                 const pad = (visible*2.5);
 
                 const bottomOfPage = (visible + scrollY) >= (pageHeight-(pageHeight>pad ? pad : 300));
 
+                const numPages =  Math.floor(scrollY/visible)+1;            // Calculate number of pages scrolled
+                var stopWordsRE = /(?:^|\s+)(?:the|The|El)(?=\s+|$)/gi;     // Pathetic attempt at stop word removal (needs to be a parameter)
```
CDrummond: I _think_ LMS might be able to list the sort-character, in which case this RE would not be required.

MaterialSkin/HTML/material/html/js/browse-page.js
```diff
@@ -1595,17 +1597,53 @@ var lmsBrowse = Vue.component("lms-browse", {
...
+                if (this.useGrid) {                                         // Grid Display, 
+                    var elem = document.getElementById("item0");
+                    var rows = Math.round( scrollY / elem.offsetHeight); 
+                    var colsPerPage  =   Math.floor(this.scrollElement.clientWidth / elem.offsetWidth) ; 
+                    var itemNum = "item"+ Math.floor(rows*colsPerPage);
+                    var classname = "v-card v-card--flat v-card--tile theme--" + ( this.$store.state.darkUi ? "dark" : "light"); 
+                    elem = document.getElementById(itemNum).getElementsByClassName(classname)[0];
+                    var  overlayHTML = elem.title.replace(stopWordsRE,"").trim().charAt(0).toUpperCase();
+		}else {                                                     // List display 
+                    var classname = "v-list__tile v-list__tile--link v-list__tile--avatar theme--" + (this.$store.state.darkUi ? "dark" : "light");
```
CDrummond: Ummm... this looks a bit odd here. Not sure I like this. why by class name?

MaterialSkin/HTML/material/html/css/desktop.css
sanc909: Good point. I will add.

MaterialSkin/HTML/material/html/js/browse-page.js
sanc909: Yes I think it can, couldn't see it how do it here so bodged a simple test.

CDrummond: Yeah, the request for artists/albums/genres would need to be updated to ask for their 'textkey' However, not sure how this would work for apps (e.g. Spotify)

CDrummond:
Whilst I like the general idea, I'm not 100% sold on it. For one, I guess it should only be active if more than X (e.g. 100?) items in a list, also it only works for alphabetically sorted lists. So, would not work for track listings, or favourites (unless the user has decided to sort).

sanc909:
Hi thanks for your comments and no problems. 
It could easily  be restricted to only action if this.listSize >= 100 ? 
Any value in me persisting?

MaterialSkin/HTML/material/html/js/browse-page.js
sanc909: So for the GridView individual items are have ids of the form "item01", "item02" etc
But from what I read, the card text showing title doesn't have an id -  so needs to be retrieved via a classname? Or am I misreading?

sanc909: No problems - thanks for your insights. Can I check my understanding on Grid view?

Each grid item has a key of item0 etc
An example title div inside the grid item is specified as 
`<div title="1 (2000)The Beatles" class="v-card v-card--flat v-card--tile theme--dark">`

I couldn't see how to access it without using the class name.  I guess solvable by generating an ID for the title div? Just want to check my understanding. thanks for your help.

sanc909:
Thanks. I can see the this might not be practical as the list partially loads, so it stalls. . I might try and get textkey working but let's close this for now.  Anyway thanks so much for your help

CDrummond:
Thought you might like to know that a modified version of this is now in master for the 0.5.0 release. Thanks for the original idea and code.

sanc909:
Hi Craig

Thanks for your note - that's great. I will take a look. Thanks for adding this and your continued excellent work on the Material skin. It's a great addition to the LMS Server


Kind Regards
San

PR #87, opened 2019-03-04
Do not submit undefined title/artist/album information as part of the…

michaelherger:
… "musicartistinfo" request

The undefined value would be stringified to "undefined", making it difficult to verify this on the server side.


CDrummond:
OK, I got impatient :-) I implemented your changes, but more in line with the rest of the code. Can you update and confirm is still works as you intended? Seems to work OK for me.

michaelherger:
I think this will be good enough for >99.99% of all cases. But I think there was an album called '0' (zero), which could potentially fail your check. But oh well... Thanks!

CDrummond:
Ah, OK - I'll change it to "x!=undefined"

PR #94, opened 2019-03-28
Fix jumplist for non-local content

michaelherger:
...which fixes the list width issue, too.

`i.filename` looks like the result of a copy/paste job. In remote services this should be `i.title`.


CDrummond:
Thanks!

Issue #100, opened 2019-04-08
Not requesting an adequately resized artwork


kayue:
I'm using the Qobuz plugin, and I noticed the image returning from image proxy is larger than the original one from Qobuz server.

Image proxy returning 1500px 

```
/imageproxy/https%3A%2F%2Fstatic.qobuz.com%2Fimages%2Fcovers%2F52%2F13%2F0094636981352_600.jpg/image.jpg
```

Original picture is 600px

```
https://static.qobuz.com/images/covers/52/13/0094636981352_600.jpg
```

Expected result would be something like this

```
/imageproxy/https%3A%2F%2Fstatic.qobuz.com%2Fimages%2Fcovers%2F52%2F13%2F0094636981352_600.jpg/image_300x300_f.png
```

The bug was originally reported at https://github.com/Logitech/slimserver/issues/247, then they lead me to here.

CDrummond:
And this is an issue because? Material uses the unscaled images for now-playing and backdrops. Why should these be scaled? Perhaps the cover in the bottom toolbar of the desktop view should be scaled. But I fail to see why it should elsewhere.

CDrummond:
http://192.168.1.16:9000/imageproxy/https%3A%2F%2Fstatic.qobuz.com%2Fimages%2Fcovers%2F52%2F13%2F0094636981352_600.jpg/image.jpg

is 300x300 for me

CDrummond:
Resolved by #101

PR #101, opened 2019-04-08
Improve artwork handling by using server-side resizing

michaelherger:
* double the size of the icons to get a crisper picture on hires displays. In a perfect world you'd load display dependent resolution using
* not only use the imageproxy for PNG files, but JPG, too
* use imageproxy for images without extension, too. For any absolute URL.

This should address the concern expressed in https://github.com/Logitech/slimserver/issues/247.


CDrummond:
Does this resolve #100 ?? To be honest, I fail to see the issue in that report...

michaelherger:
Ah yes, same report, same confusing story :-).

To put it short: in some cases you'd just pull in the original image, which an be pretty heavy. Some services provide artwork which weighs > 1MB per picture. This times the number of images on a page can become heavy. Resizing on the server doesn't decrease the bandwidth used, but at least caches stuff for other devices. Eg. you render the list on the phone, and later on the tablet. LMS would serve quickly from its cached the resized image.

As for the quality increase, there's this for eg. icons:

Bildschirmfoto 2019-04-08 um 18 12 25
Bildschirmfoto 2019-04-08 um 18 13 18

PR #103, opened 2019-04-10
German updates

michaelherger:


CDrummond:
That was quick!

michaelherger:
Coincidence: I thought I'd see what was there. And while I added my three lines, you updated on your end, too :-)

PR #115, opened 2019-04-30
Display the git revision when running from a git clone

michaelherger:
Giving a time stamp is pretty useless, as it would always be the time when the page was loaded, without any reference to what we're actually running.

My proposed implementation admittedly is reaching out deep into LMS code... but it should allow us to reliably figure out where the plugin is being installed, as it can be in quite a few different places. It does get the list of folders being used by LMS' templating engine, and picks the one matching the MaterialSkin name. That folder can be expected to be in the git clone, where we can get the revision.

### DIFF
--- a/MaterialSkin/Plugin.pm
+++ b/MaterialSkin/Plugin.pm
@@ -20,11 +20,23 @@ sub initPlugin {
 sub pluginVersion {
     my ($class) = @_;
     my $version = Slim::Utils::PluginManager->dataForPlugin($class)->{version};
+    
+    if ($version eq 'DEVELOPMENT') {
+        # Try to get the git revision from which we're running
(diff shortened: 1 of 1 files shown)


CDrummond:
Using time is not "useless" - as it works for my scenario. All I care about here is making sure the non-cached version is used. I agree its not as useful as git revision - as this will only cause the cache to change when there is a change - but need git be installed, and the plugin to have access to its .git folder. On my setup the .git is not transferred, so this would not work.

This change should check whether getting the git tag succeeds, and if not then fall back to using the time as I have done. This way it would work for all - and use the better (git) method in preference.

michaelherger:
Heh... I had this fallback in place, but removed it, as I considered the timestamp useless. But for you yourself I guess it's indeed good enough. For external users, though, I'd expect them to be using git to test your dev code anyway. Therefore in those cases the git revision would be more helpful, as we users could tell you easily what was running.

I'll push a new change to fall back to your version.

PR #132, opened 2019-07-24
Make sure the scanner does pre-cache album artwork in the size used i…

michaelherger:
…n browse modes

Without pre-caching LMS would resize images on the fly. This can considerably slow down browsing the library. By pre-caching those files LMS can quickly read the re-sized artwork from its cache.

Make sure the spec matches the LMS_GRID_IMAGE_SIZE in constants.js. 

iPeng users might already have this size registered. But it doesn't hurt to configure it for the other users, too. See Settings/Advanced/Performance.

### DIFF
--- a/MaterialSkin/Plugin.pm
+++ b/MaterialSkin/Plugin.pm
@@ -43,6 +43,9 @@ sub initPlugin {
             return Slim::Web::HTTP::filltemplatefile('mobile.html', $params);
         } );
         Slim::Web::Pages->addRawFunction($URL_PARSER_RE, \&_svgHandler);
+
+        # make sure scanner does pre-cache artwork in the size the skin is using in browse modesl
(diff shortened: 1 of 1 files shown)


CDrummond:
...and for list size?

michaelherger:
Aren’t you using the same image size in lists, but let the browser resize them client side (through some CSS rule)?

CDrummond:
Nope. I'm using 100x100_o - would it be better to just let the browse resize? I'm leaning towards this - less code :-)

CDrummond:
Changed co de to use 300x300 everywhere, and have browser resize.

michaelherger:
Heh... I would have bet I did double check that... but it's been on the train, around midnight, after a nice dinner on the lake side... new PR coming.

michaelherger:
NM - just saw your latest commits.

PR #133, opened 2019-07-25
Update fr.json

Franck-Berry:
Following our discussion yesterday on the LMS forum (my nickname on the forum is 'Franck@'), here is a small evolution of the French translation. (I created a GitHub account in between.)


CDrummond:
Thanks!

PR #170, opened 2019-10-10
Some more tweaks to the text rendering for the settings pages.

michaelherger:
* make sure we're using the same Roboto font everywhere
* apply font smoothing
* lighten some of the controls
* use pointer cursor for controls

Tested with Firefox, Chrome, and Safari on macOS


michaelherger:
Old vs. new shows the font no being that bold, much closer to the rest of the Material skin:
Bildschirmfoto 2019-10-11 um 00 47 43
Bildschirmfoto 2019-10-11 um 00 47 21
Bildschirmfoto 2019-10-11 um 00 48 09

PR #172, opened 2019-10-13
Fix #171 - Use light/dark skin in settings depending on Material pref

### LINKED ISSUE #171: Light Material skin would still load settings in dark mode
The dark stylesheet is hard-coded. If you could pass the iframe the skin as a parameter I could add a few lines on the server side to make it work. But I don't know how you're calling that iframe with the settings.

michaelherger:
* add parameter "darkUi" to calling the settings
* add standardheader.html template override to handle the skin selection
* rename slimserver.css to skin.css, no need to Classic's slimserver.css
* tweak some styles for iOS devices
  - need to reset button appearance in order to style them
  - give selects a 3px radius


CDrummond:
Thanks for this!

PR #173, opened 2019-10-13
Remove some testing leftover...

michaelherger:


CDrummond:
Thanks again. I sent you a message on LMS forums - basically, the 3px radius added to selects, I've removed this (again) as Material does not have these. Is it **required** for iOS? Any rendering errors without? Also, I re-added the styling of buttons - as Material has a light shadow for non-flat buttons (and the CSS attempts to mimic this).

PR #218, opened 2020-01-27
Expand names and descriptions in web app manifests

mavit:
I find three letter acronyms unfriendly.  There's not enough space in the `short_name` field to spell the name out in full, but there is in the `name` field.  I realise this is subjective, so won't be offended if this is rejected.

While here I've added a category and a description.

I've also added a `.editorconfig` file which will cause compatible editors to indent .webmanifest files by 2 columns.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.webmanifest]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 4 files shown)


CDrummond:
Dont like all the changes, plus please don't add editor settings - as I'm pretty sur eI use a different edtiro to you. I have applied some of the changes, with credit to you.

mavit:
> I'm pretty sure I use a different editor to you.

I expect so, but the idea of [EditorConfig](https://editorconfig.org/) is that’s it’s supported by lots of different editors.  Even if yours doesn't, I would make it easier for other contributors to indent their code the way you like it.

>I have applied some of the changes, with credit to you.

Thanks!

PR #256, opened 2020-03-31
`use strict` - and fix some undefined variables

michaelherger:
You should always `use strict`, as this will tell you about undefined variables etc. I'm surprised you didn't see any crashes with your code, as it did use undefined variables in a few places. Others I'm not sure why they would work, as they are being used in an always undefined state... 

I also changed to use `cstring($client, ...)`, as `cstring()` is specifically there to return client specific translations. If you don't want them, then you could just use `string()` instead.

### DIFF
--- a/MaterialSkin/Plugin.pm
+++ b/MaterialSkin/Plugin.pm
@@ -8,6 +8,8 @@ package Plugins::MaterialSkin::Plugin;
 # MIT license.
 #
 
+use strict;
+
 use Config;
 use Slim::Menu::BrowseLibrary;
 use Slim::Music::VirtualLibraries;
@@ -92,7 +94,7 @@ sub initPlugin {
 sub pluginVersion {
     my ($class) = @_;
     my $version = Slim::Utils::PluginManager->dataForPlugin($class)->{version};
-    
+
     if ($version eq 'DEVELOPMENT') {
         # Try to get the git revision from which we're running
         if (my ($skinDir) = grep /MaterialSkin/, @{Slim::Web::HTTP::getSkinManager()->_getSkinDirs() || []}) {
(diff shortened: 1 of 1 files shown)


MaterialSkin/Plugin.pm
```diff
@@ -186,17 +191,17 @@ sub _cliCommand {
         my $osDetails = Slim::Utils::OSDetect::details();
         my $serverPrefs = preferences('server');
         $request->addResult('info', '{"server":'
-                                .'[ {"label":"' . cstring('', 'INFORMATION_VERSION') . '", "text":"' . $::VERSION . ' - ' . $::REVISION . ' @ ' . $::BUILDDATE . '"},'
-                                .  '{"label":"' . cstring('', 'INFORMATION_HOSTNAME') . '", "text":"' . Slim::Utils::Network::hostName() . '"},'
-                                .  '{"label":"' . cstring('', 'INFORMATION_SERVER_IP') . '", "text":"' . Slim::Utils::Network::serverAddr() . '"},'
-                                .  '{"label":"' . cstring('', 'INFORMATION_OPERATINGSYSTEM') . '", "text":"' . $osDetails->{'osName'} . ' - ' . $serverPrefs->get('language') .
+                                .'[ {"label":"' . cstring($client, 'INFORMATION_VERSION') . '", "text":"' . $::VERSION . ' - ' . $::REVISION . ' @ ' . $::BUILDDATE . '"},'
```
CDrummond: What if there is no client? For this 'info' command no client is specified.

michaelherger: That’s what cstring() is for: it checks, whether a client is defined or not. If you know there’s none, then just use string() instead.

PR #280, opened 2020-05-18
Update of the French translation.

Franck-Berry:
For information, the file fr.json does not give the possibility of translating the name of theme "Windows 10 Dark" (Windows 10 Sombre).
Similarly, when you restart the server, the buttons "yes" and "no" do not seem to be translatable into "oui" and "non".

### DIFF
--- a/MaterialSkin/HTML/material/html/lang/fr.json
+++ b/MaterialSkin/HTML/material/html/lang/fr.json
@@ -446,9 +446,9 @@
 "%1 Years":"%1 années",
 "1 day":"1 jour",
 "%1 days":"%1 jours",
-"Color toolbars":"",
-"Darker":"",
-"Show current date and time":"",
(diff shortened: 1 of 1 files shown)


CDrummond:
'Windows 10 Dark' comes from the filename. I'll see about making the 'Dark' part translatable.

Not sure what yes/no buttons you are referring to. In "Server Settings" if I click "Restart" I get "Restart" and "Cancel"

CDrummond:
For 1.6.1 I've updated the code so that the "Dark" part of the name will be translated using the existing "Dark" translation.

Franck-Berry:
Thank you for updating the theme name.

For the "yes" and "no" buttons, it is not when the server is restarted. I thought, however. Sorry to have given you false information. No way to remember in what context these buttons appeared. Next time, I'll take a screenshot.

Thank you again for your responsiveness and for the skin!

PR #284, opened 2020-05-24
Tolerate non-string firmware values

UrbaneChimp:
Player drop-down creation was failing because some of my players have firmware values that are numbers and not strings.


CDrummond:
Thanks!

PR #294, opened 2020-06-12
Danish translation update

svalgaard:
Danish translation update including using the same word for "alarms" as used in the official LMS translation.


CDrummond:
Thanks!

PR #309, opened 2020-08-05
Fix URL query handling of urlencoded player names

hwhesselink:
Player names are passed in as urlencoded strings (.../material/?player=Player%20Name) but weren't being decoded.  This patch adds a call to decodeURIComponent when fetching the parameter.


CDrummond:
Thanks!

PR #377, opened 2020-11-22
Optimized macOS desktop image

michaelherger:
Here's something that really, really bothered me for a long time 😁: the desktop icon for Material skin used by an "appified" Chrome instance is too big.

Bildschirmfoto 2020-11-22 um 12 16 20

According to [Apple style guides[(https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/app-icon/) you should have some whitespace (gutter?) around your icon. I'd therefore suggest you replace the existing icon with my copy of it (which I created based on your svg file):

Bildschirmfoto 2020-11-22 um 12 32 37

The first icon is yours, second is my modified version, next to other popular circular icons.


CDrummond:
Thanks, but the circle is now smaller than safari and chrome. Can you resize so that it is the same? I'd do this myself but I don't have a mac. I think its about 4px (so 2px either side) smaller.

michaelherger:
Wow... is it? I'll give it another try. Damned pixel peeper 😁!

CDrummond:
Yeah, sorry - I can be a real pain in the proverbial!

michaelherger:
Did you actually measure the screenshot, or did you spot this like that?

michaelherger:
Bildschirmfoto 2020-11-22 um 14 27 43

CDrummond:
Looks better now. I did notice that it _looked_ slightly smaller by eye, and then I placed another window over the screenshot to detect the height.

PR #400, opened 2021-01-12
Un-squeeze treble clef

michaelherger:
Probably more of a kind nudge than a PR worth the pull, as I'm not sure how you edit your SVG files. I used whatever I found...

treble-clef-before-after

The treble clef previously was a bit squeezed (right hand side). My version (on the left) would un-squeeze it a bit.


CDrummond:
The "fattening" of the icon was deliberate. IMHO, your version looks to thin compare to the other icons in the "My Music" list.

I used inkscape to stretch it a bit, and then 'scour' to trim the SVG contents to the bare minimum.

So, not sure about this. I tried your icon, but I still think a thicker (or wider) version looks better. Perhaps if the icon itself was of those proportions, rather than me crudely stretching it, then it would look better. But, I'm no inkscape expert or icon designer.

michaelherger:
Agreed, it looks very thin. But you version always looked wrong to me 😁. Maybe the treble clef just isn't suitable here. As a former trombone player I'd suggest the bass clef instead 😉.

Qobuz uses something I kind of like. Don't know whether it would fit in well. But for those who remember disc stores...

qobuz-file

CDrummond:
I've changed the genres icon to an acoustic guitar. Does this look OK?

michaelherger:
That would work, too. The image looks cropped, though. Top right it seems to be missing a few details.

And I guess you'll need to update a few more mappings. Eg. Spotty's Genre item still features the treble key.

Thanks for all your efforts!

CDrummond:
The cropping is from the original icon - its the "style". I _think_ its meant to look like the 'head' part is leaning back.  I obtained the SVG from materialdesignicons.com

CDrummond:
Closing. I've now removed this icon, and used a better icon for the guitar.

michaelherger:
Thanks!

PR #414, opened 2021-01-29
Further BBC Sounds Icon mapping

expectingtofly:
Many thanks for mapping the BBC Sounds Icons.
I've mapped further icons that are in the deeper menus and made a couple of changes to closer match the BBC's icons.

image

image

image

image


CDrummond:
Thanks for this. Could the BBC sounds icons not be given more generic names? e.g. BBCSoundsSearch.png -> search.png? That way icon mapping will apply to other apps to. It'd be nice if LMS provided a bunch of generic icons, and then Material would map those. What I'm waeary of is each new plugin having (e.g.) 20 plugin specific names for icons - which means Material has more processing to do.

Not a massive issue, but would be nicer.

expectingtofly:
> Thanks for this. Could the BBC sounds icons not be given more generic names? e.g. BBCSoundsSearch.png -> search.png? That way icon mapping will apply to other apps to. It'd be nice if LMS provided a bunch of generic icons, and then Material would map those. What I'm waeary of is each new plugin having (e.g.) 20 plugin specific names for icons - which means Material has more processing to do.
> 
> Not a massive issue, but would be nicer.

Yes, agreed.   Maybe the answer is to have a naming standard that you support so Material can automatically pick the svg icon.
something like MTL_[matrial standard name].svg.   So "MTL_radio-tower.svg" would automatically get mapped to the radio-tower in the svg library?

I'm happy to make that change now rather than the hard coded mapping if you're happy with that?

CDrummond:
Not sure it needs to be so meterial specific - after all other UIs may wish to map icons. Just that, if possible, the names should be generic - and match their purpose.

CDrummond:
So search.png, subscribe.png, latest.png, tracks.png, subscriptions.png, mymusic.png, etc.

CDrummond:
...those were just examples, if you can think of better, then fine. Just not 'BBCSoundsxxxx.png' ;)

expectingtofly:
> So search.png, subscribe.png, latest.png, tracks.png, subscriptions.png, mymusic.png, etc.

Yes, I understood.  I was just trying to think of better solution. After all, these are BBC Sounds specific icons.  We just need to map onto the Material specific icons so that Material theming can occur.  

Your suggested solution is to use generic names as the convention so that the mapping can occur, I'm suggesting we could put the onus on the plugin developer to use a naming convention so that Material could automatically map to its own icons that are already in its library.   That way you don't need to continually add in the mappings.

I'm not aware of any other controller that tries to do icon theming, so it would be transparent to everything else, and even if there was one, it wouldn't matter.

I've done a code change to my fork of material that looks for MTL_icon[name]_MTL_end in the image name and maps it into the library eg : 
"folder/mymusic_MTL_icon_recent_actors_MTL_end.png"   would be mapped onto the "recent_actors" svg in the material library.

If you are interested in looking at that pull request I'll send it over?

CDrummond:
Yeah, I'd be interested. I'm just weary of an tying things to any specific UI, etc.

expectingtofly:
Agreed that hard coding of plugin specific icon mapping is bad and an unnecessary workload on the maintainer.  We'll think of another way.

PR #415, opened 2021-01-29
Generic mapping from plugin icons to material library icons through naming conventions

expectingtofly:
As part of the icon mapping, adds in a check for a naming convention within the image name so that the plugin specific icon can be automatically mapped onto the Material icon library.  Allowing an svg to be used and themed.
This puts the onus on the plugin developer to make their plugin compatible with Material theming rather than a change to Material each time.  It also allows the plugin developer to choose which Material icon is used.

The code looks for 'MTL_icon' or 'MTL_svg' within the image name and extract the string up to '.'.  It then uses that as the icon library reference.

e.g. 
The image  /html/img/Speech_MTL_icon_format_quote.png

would be mapped to the 'format_quote'  material library icon.


CDrummond:
I'm happy with this, thanks! But, is _MTL_end required? Why not just remove ".png"? Seems a bit superfluous. So, `/html/img/Speech_MTL_icon_format_quote.png` would suffice.

CDrummond:
...or look for the last period - therefore allowing different extensions.

expectingtofly:
> ...or look for the last period - therefore allowing different extensions.

yeah,  I had it like that first.  And tested it like that.  So just replacing the '_MTL_end' with a '.' in the code works.
Just thought some might not want to be forced to put the material icon name at the end.
But I'm happy either way, whatever you think best.

CDrummond:
I though the start was the plugin specific part. Seeing as after MTL_ is material, I can't imagine there will be other UI specified here - ie not Wibble_MTL_icon_search_MTL_end_iPeng_icon_wibble_iPeng_end, etc, in which case just Wibble_MTL_icon_search.png suffices - and, to me, looks cleaner than Wibble_MTL_icon_search_MTL_end.png.

expectingtofly:
> I though the start was the plugin specific part. Seeing as after MTL_ is material, I can't imagine there will be other UI specified here - ie not Wibble_MTL_icon_search_MTL_end_iPeng_icon_wibble_iPeng_end, etc, in which case just Wibble_MTL_icon_search.png suffices - and, to me, looks cleaner than Wibble_MTL_icon_search_MTL_end.png.

Fine by me.  I'll push another commit

expectingtofly:
Branch has been updated

PR #416, opened 2021-01-30
Update nl.json

Blackfiction:
Updated nl strings. Please note a typo in the english string; -> 'pinnnig' -> pinning
If you store the library when pinnnig then this library will always be used


CDrummond:
Thanks! ...and for spotting the typo - fixed now.

PR #432, opened 2021-02-22
Use LMS' own advanced search params parser (if available - requires 8.2+)

michaelherger:
If LMS supports the `Slim::Web::Pages::Search::parseAdvancedSearchParams()` function, use it instead of the plugin's copy of that code.

In order to achieve this I've added a stub inline module `Plugins::MaterialSkin::Search`. If LMS doesn't support the aforementioned call, the plugin is requiring the `Search.pm` module, which would override the stub with its own implementation.

New LMS builds supporting this feature should be out soon, the change already is committed to https://github.com/Logitech/slimserver/commit/d776dc6292195e6511cb14c741bd39049b872614.

### DIFF
--- a/MaterialSkin/Plugin.pm
+++ b/MaterialSkin/Plugin.pm
@@ -23,7 +23,10 @@ use Slim::Utils::Strings qw(string cstring);
 use HTTP::Status qw(RC_NOT_FOUND RC_OK);
 use File::Basename;
 use File::Slurp qw(read_file);
-use Plugins::MaterialSkin::Search;
+
+if (!Slim::Web::Pages::Search->can('parseAdvancedSearchParams')) {
+    require Plugins::MaterialSkin::Search;
+}
 
 my $log = Slim::Utils::Log->addLogCategory({
     'category' => 'plugin.material-skin',
@@ -830,7 +833,7 @@ sub _cliCommand {
             $params->{'saveSearch'} = $saveLib;
         }
 
-        my ($tracks, $albums) = Plugins::MaterialSkin::Search::advancedSearch($params);
+        my ($tracks, $albums) = Plugins::MaterialSkin::Search::advancedSearch($request->client(), $params);
 
         if (!$saveLib) {
             if ($tracks) {
@@ -904,7 +907,8 @@ sub _cliCommand {
     }
 
     if ($cmd eq 'adv-search-params') {
-        my $params = Plugins::MaterialSkin::Search::options();
+        my $params = Plugins::MaterialSkin::Search::options($request->client());
+
         my $count = 0;
         while (my $genre = $params->{'genres'}->next) {
             $request->addResultLoop('genres_loop', $count, 'id', $genre->id);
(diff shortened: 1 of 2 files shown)


MaterialSkin/Plugin.pm
```diff
@@ -830,7 +833,7 @@ sub _cliCommand {
             $params->{'saveSearch'} = $saveLib;
         }
 
-        my ($tracks, $albums) = Plugins::MaterialSkin::Search::advancedSearch($params);
+        my ($tracks, $albums) = Plugins::MaterialSkin::Search::advancedSearch($request->client(), $params);
```
CDrummond, line 836: Why is a client required? Material does not passa client into this API, as it is not used.

MaterialSkin/Plugin.pm
```diff
@@ -904,7 +907,8 @@ sub _cliCommand {
     }
 
     if ($cmd eq 'adv-search-params') {
-        my $params = Plugins::MaterialSkin::Search::options();
+        my $params = Plugins::MaterialSkin::Search::options($request->client());
```
CDrummond, line 910: Again, no client.

MaterialSkin/Plugin.pm
```diff
@@ -1245,3 +1249,21 @@ sub _userColorHandler {
...
+
+# Small stub for Plugins::MaterialSkin::Search to use LMS methods if available
+package Plugins::MaterialSkin::Search;
+
+use Slim::Web::Pages::Search;
+
+sub advancedSearch {
+    my ($client, $params) = @_;
+
+    $params->{'searchType'} ||= 'Album';
```
CDrummond, line 1261: Material does not set searchType, as it always wants both tracks and albums. This looks like its defaulting to Albums.

MaterialSkin/Search.pm
```diff
@@ -25,7 +25,7 @@ use Slim::Utils::Log;
 use Slim::Utils::Prefs;
 
 sub advancedSearch {
-	my $params  = shift;
+	my ($client, $params) = @_;
```
CDrummond, line 28: Why? `client` is not used.

michaelherger: Your implementation does not use it, but LMS' does. By keeping the function's signature identical, we can simplify the calling side.

Feel free to replace `$client` with `undef` if you don't like to have an unused variable defined.

MaterialSkin/Plugin.pm
michaelherger: You commented out condition in your copy of the implementation. But as LMS is expecting it, we inject it here. Like with the `$client` parameter I'm doing this to keep the interface between the two implementations identical.

CDrummond: But will this not restrict the response to albums?

MaterialSkin/Search.pm
CDrummond: Fair enough :)

MaterialSkin/Plugin.pm
CDrummond: ...ah, looking at the LMS code it needs this to **also** get albums. My bad.

PR #509, opened 2021-07-25
Update ru.json

sky418:
587 - Small 
In what context is this expression used  - in the sense of action (make something small ) or in the literal sense (small font ....) 
There are Different translation in different cases


CDrummond:
"Small" refers to the font-size setting; Large, Regular, Small

sky418:
THX. So, everuthing OK

PR #544, opened 2021-12-05
czo-sync-menu

czodroid:
Put 2 options : "Show the manage players menu" and "Show the synchronise menu"
It's like always for people who don't act on these two options. But for me, it's great: I can set "Show management drives" to OFF and "Show sync menu" to ON.
I have a lot of squeezeboxes, 1 Transporter, 1 Classic, 1 Boom, 3 Radio and my 2 laptops that use squeezelite.
I'm 52, and presbyopic, so on my phone, with a large font size, it's really hard to see all my squeezeboxes in the "manage drives" dialog, so I prefer to have the synchronise drop down menu for that.


CDrummond:
Not sure I really want more options just for this. Perhaps, if the screen is too narrow the Manage players dialog should not show the current cover - this will give more space for the player name, and should resolve your issue with that.

CDrummond:
As stated, I don't want superlous config items to control menu entries. I have adjusted the sync dialog to give more space for player names, adjusted 'Manage players' so that more space is given for names when using narrow screen widths. I have also added access to the sync dialog from within the players settings dialog - as a button near the bottom, and an entry within its own menu (top right of dialog).

czodroid:
Hi,

I'm so so sorry !!!!

I liked lms v1.6 because of the "synchronise dialog", and posted this on
forums.slimdevices.com on 6/4/2020 at 18:58.
In the "player menu", I still prefer this "synchronise dialog" instead of
the "manage players".

Maybe I'm not alone.

There are more than 30 options in the Interface section, so it's not bad to
add 2 more options !!!

Please, Please, re-open the ticket !

In the "manage players", it is difficult to (un)-synchronize a player, on a
small smartphone....But with your "synchronise dialog" it's perfect for me:
with it you can select all or none in two key presses !!! See the jpg
attached in this email.

Thank you - as always -  for LMS !

BR,

Olivier.

PR #601, opened 2022-07-06
Update nl.json

Blackfiction:


CDrummond:
Thanks. You can do 1 pull request for both strings.txt and this, no for separate requests. Not an issue either way, whichever is easier for you, but just thought I'd mention.

PR #620, opened 2022-11-21
Auto light/dark mode

mavit:
I've been having a play around with detecting whether the user has their interface set to light or dark mode, and applying that setting to LMS Material.

I can work on this further if you think this could be a feature you'd be interested in?  I'm imagining adding an additional theme option, "Auto", and making it the default.


CDrummond:
A version of this has now been merged in to the dev branch. Thanks.

mavit:
Thanks.  Apologies that I never got around to finishing it.

PR #713, opened 2023-09-10
add icon to favorite editor

wizmo2:
added ability to edit the icon.  When adding http links, they are converted to use `imageproxy`.

This is a feature I just happened to need, thought I'd share in case you wanted to add it.

There are limitations associated with LMS (png,jpg,gif formats only).

### DIFF
--- a/MaterialSkin/HTML/material/html/js/favorite-dialog.js
+++ b/MaterialSkin/HTML/material/html/js/favorite-dialog.js
@@ -23,6 +23,11 @@ Vue.component('lms-favorite', {
       <v-text-field clearable autocorrect="off" :label="i18n('URL')" v-model="url" class="lms-search"></v-text-field>
      </v-list-tile-content>
     </v-list-tile>
+    <v-list-tile>
+     <v-list-tile-content>
+      <v-text-field clearable autocorrect="off" :label="i18n('Icon')" v-model="icon" class="lms-search"></v-text-field>
+     </v-list-tile-content>
+    </v-list-tile>
    </v-list>
   </v-form>
   <v-card-actions v-if="queryParams.altBtnLayout">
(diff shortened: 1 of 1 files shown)


MaterialSkin/HTML/material/html/js/favorite-dialog.js
```diff
@@ -103,8 +113,12 @@ Vue.component('lms-favorite', {
                 lmsCommand(this.playerId, ["favorites", "delete", this.item.id]).then(({data})=> {
                     var command = ["favorites", "add", "url:"+url, "title:"+name, this.item.id];
                     if (this.item.presetParams) {
-                        if (this.item.presetParams.icon) {
-                            command.push("icon:"+this.item.presetParams.icon);
+                        if (icon) {
+                            if (icon.startsWith("http")) {
```
CDrummond, line 117: Why are you doing this here? If I favourite from Planet Radio, or Bandcamp, etc. the icon starts with "http://"if I now edit the favourite to change its title this will then also edit the image URL - which would be unexpected.

wizmo2: IMO, forcing image proxy is a better solution.  Saying that, I can look at adding a checkbox.

P.S.  In the scenario of Name change only, the icon is not updated/

CDrummond: OK, ,my mistake. If I change the URL then the icon is also modified.

No, a checkbox to auto-add imageproxy makes no sense. Why do you want the proxy added to the URL? Material shold be adding it itself anyway.

CDrummond: Sorry if it seems I'm being difficult with this - I do appreciate your efforts and I'm more than happy to merge this for 3.5.0 (with the fixes above).

PR #731, opened 2023-09-27
Adjust album grid paddings

denys-lysiuk:
Hi Craig!
The paddings around album items in grid view always looked a little bit off to me being uneven on left and right edges, and in-between, so I've tried to apply a small fix and want to introduce it to you. Please take a look if it makes sense. And thanks for your work btw.

Narrow view - before:
narrow-before

Narrow view - after:
narrow-after

Standard view - before:
standard-before

Standard view - after:
standard-after


CDrummond:
Wow, great! Thanks for this.

PR #747, opened 2023-10-17
Pre-cache HiDPI version of the cover arts if needed

michaelherger:
The pref is set by LMS when it sees a HiDPI capable display. You'd currently "only" pre-cache 300px, but not 600px, slowing down the scrolling on slower servers.


CDrummond:
Thanks!

PR #772, opened 2023-11-13
Fix version check.

michaelherger:
The current check is broken, as it's comparing against he first part of the version string even for the second digit.

Use LMS built-in `compareVersions()` instead.


CDrummond:
Ah, oops! Thanks!

PR #846, opened 2024-02-26
Reduce rather than clip icons in Favorites list

bert-github:
I have radio stations in my list of favorites, but their icons are clipped, which makes them hard to recognize and doesn't look good. In the list of radio stations, on the other hand, they are reduced to fit, which looks much better. I propose this fix to the CSS rules for the favorites list: In two places change 'object-fit: cover' to 'object-fit: contain'; and remove the 'object-position', so the icon is centered rather than aligned at the top.

### DIFF
--- a/MaterialSkin/HTML/material/html/css/style.css
+++ b/MaterialSkin/HTML/material/html/css/style.css
@@ -542,8 +542,7 @@ div.v-subheader {
 }
 
 .image-grid-item-img {
- object-fit:cover;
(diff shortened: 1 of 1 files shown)


CDrummond:
Sorry, but no. this change would affect **all** grid/lists - and this clipping is added for a reason. However, your issue does have merit so I've added #848 to look into fixing this.

However, thanks for taking the time to look for a solution.

PR #875, opened 2024-04-22
Update nl.json

Cadsters:
Better dutch translation for works


Cadsters:
In the context of work it's the wrong translation, i will commit an update.

PR #877, opened 2024-04-22
Title display for works (draft)

darrell-k:
PR created just as an aid to discussion.

The issue this is trying to solve is to include the Work in the Track Title, but only if the Track Title doesn't already contain the Work. The background is that people (like me!) can have a mixture of tagging for classical works, as we have struggled to make something work for LMS (before the recent enhancements) and other servers. (That's without considering the inconsistencies in the tagging of tracks from online services).

For example, a classical track could be tagged either:
WORK="Violin Concerto in D Major Op.77"
TITLE="II. Adagio"
or
WORK="Violin Concerto in D Major Op.77"
TITLE="Violin Concerto in D Major Op.77: II. Adagio"

In the first case we need to concatenate the WORK and TITLE into the displayed track title, otherwise it makes no sense. In the second case we do not want to do this.

This PR demonstrates what I mean, in the playlist display. I have lower-cased and stripped non-alphanumeric characters out of the test, so that it's not thrown off course by minor typo differences (I do something similar in the Qobuz plugin).

Whatever change is made there will also need to apply to the browse listing, now playing and the central column of the (i) information (MAI?) display.

Thanks,
Darrell

### DIFF
--- a/MaterialSkin/HTML/material/html/js/queue-page.js
+++ b/MaterialSkin/HTML/material/html/js/queue-page.js
@@ -94,7 +94,7 @@ function parseResp(data, showTrackNum, index, showRatings, queueAlbumStyle, queu
             for (var idx=0, loop=data.result.playlist_loop, loopLen=loop.length; idx<loopLen; ++idx) {
                 let i = loop[idx];
                 splitMultiples(i);
-                let title = i.title;
+                let title = trackTitle(i);
                 if (showTrackNum && i.tracknum>0) {
                     title = formatTrackNum(i)+SEPARATOR+title;
                 }
@@ -456,7 +456,7 @@ var lmsQueue = Vue.component("lms-queue", {
                 var index = playerStatus.current["playlist index"];
                 if (undefined!=index && index>=0 && index<this.items.length) {
                     var i = playerStatus.current;
-                    var title = i.title;
+                    var title = trackTitle(i);
                     var rating = this.$store.state.showRating && undefined!=i.rating ? Math.ceil(i.rating/10.0)/2.0 : undefined;
                     if (this.$store.state.queueShowTrackNum && i.tracknum>0) {
                         title = formatTrackNum(i)+SEPARATOR+title;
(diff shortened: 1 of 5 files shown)


MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -94,7 +94,7 @@ function parseResp(data, showTrackNum, index, showRatings, queueAlbumStyle, queu
             for (var idx=0, loop=data.result.playlist_loop, loopLen=loop.length; idx<loopLen; ++idx) {
                 let i = loop[idx];
                 splitMultiples(i);
-                let title = i.title;
+                let title = i.title.toLowerCase().replace(/\W/g, '').includes(i.work.toLowerCase().replace(/\W/g, '')) ? i.title : i.work+SEPARATOR+i.title;
```
CDrummond, line 97: This should really be:
```javascript
let title = undefined==i.work || i.title.toLowerCase().replace(/\W/g, '').includes(i.work.toLowerCase().replace(/\W/g, '')) ? i.title : i.work+SEPARATOR+i.title;
```

...as only want to do this for items that have work - no point in performing the lowercase conversion, etc., if there is no work.

Also, would be better to add this to a function - perhaps in `utils.js` e.g.

```javascript
function trackTitle(i) {
    return undefined==i.work || i.title.toLowerCase().replace(/\W/g, '').includes(i.work.toLowerCase().replace(/\W/g, '')) ? i.title : i.work+SEPARATOR+i.title;
}
```

darrell-k: Yes, after creating this PR I found  I had to put in the defined check, otherwise the JS can error with work not defined.

The function idea is a good one, and I'll try to find where to insert the call for now playing, etc, but it might take me some time. I'd be more than happy (and grateful) if you could take this on! :)

CDrummond:
I've now updated master for this - queue, now-playing, and browse. Can you please confirm (seems to work for me), and close this pull request if so.

Also, can you check if the other issues you opened are in fact fixed.

darrell-k:
Yes it's fine - thanks.

PR #897, opened 2024-06-03
Replace grouping parameter with performance parameter

darrell-k:
I hope I haven't replaced `grouping` in places where it was used for something other than the `tracks.grouping` value! (definitely worth checking)

Corresponding PR for LMS is https://github.com/LMS-Community/slimserver/pull/1109: we should try to co-ordinate release. Whether this means backporting this to master, I don't know, depends on your schedule for merging current dev.

Once we've got this done, we can look at reusing `grouping` for grouping tracks into Acts, Scenes, etc in the tracks listings.


darrell-k:
Oops, created the PR against master - I'll close and try again.

PR #900, opened 2024-06-05
Performance fixes

darrell-k:
This makes sure the `performance` parameter is passed to the `tracks` command even if it is null, because null is meaningful (you may have an album where one version of the work has performance populated and a second one doesn't).

It also sets performance to -1 to indicate all performances (null or not) when browsing "all tracks".

It requires the latest https://github.com/LMS-Community/slimserver/pull/1109 to work properly. (This is now ready to go and also includes the fix to return `work_id` in the status command response.)

PLEASE make sure I've got this right and haven't broken anything else.

BTW, in testing I found that adding all tracks to the playlist using the buttons in the header is adding all tracks in the library when used from a Works--All Tracks list. This  happens without these changes, but I wouldn't have a clue where to start looking for the bug: JSON request is:
JSON REQ: ["00:04:20:12:4b:74",["playlistcontrol","cmd:add","currentaction:3"]]


darrell-k:
Done it again - wrong base. Closing this one will recreate.

PR #915, opened 2024-06-29
Adding new option to hide some players

VirtuousRivality:
This is a proposal to add an option enabling users to hide players from the list. It would be based on players to hide having name starting with '!'; this will let users who want to use that option to rename players to hide, prefixing them with '!', and enable this option.

My use case is that I use an old Squeezebox Touch, but only to control other players; it is still showing in the list of players while it is not connected to any audio output, thus I would like to be able to hide it. I tested that code locally and it seems to work fine, though I will admit I am unsure if the way I implemented this is the most elegant one.

### DIFF
--- a/MaterialSkin/HTML/EN/plugins/MaterialSkin/settings/basic.html
+++ b/MaterialSkin/HTML/EN/plugins/MaterialSkin/settings/basic.html
@@ -28,6 +28,10 @@
     <input type="checkbox" name="pref_combineAppsAndRadio" id="pref_combineAppsAndRadio" [% IF prefs.pref_combineAppsAndRadio %] checked [% END %] value="1"/>
     [% END %]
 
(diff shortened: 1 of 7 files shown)


CDrummond:
Sorry, not keen on having this exposed so much, for a very niche use case. However, I have (just now) made a change to allow hiding players by adding an entry to the preference file - see https://github.com/CDrummond/lms-material/wiki/14-Extra-configuration-items

CDrummond:
...but thanks so much for taking the time to implement this yourself!

VirtuousRivality:
No problem, I am glad you came with an option that does cover my need without making it too bold given it is indeed a niche use case :) thank you for having taken time to consider this and not simply discard the idea entirely. I am not always having a lot of time, but if there is anything you need some help with in terms of developping Material Skin further, please do not hesitate to ask me, I will see what I can do.

PR #955, opened 2024-09-27
Fixes for track highlighting

darrell-k:
This works for me...


CDrummond:
Seems to work for me too. Not sure why I had the comment about not being able to supply role_id if artist_id is removed, as this is what this change does (can have role_id without artist_id).

PR #974, opened 2024-11-03
Favourited classical album-works - some rework

darrell-k:
This seems to work...


CDrummond:
Looks good to me, thanks!

Issue #992, opened 2024-11-27
Works grouping for play queue


darrell-k:
Could the toggled version of the play queue group by Work within Album?

CDrummond:
Implemented in master using work_id + performance. Seems to work for me, can you test?

darrell-k:
Yes, it's working, but could we incorporate the work name into the title (maybe before the Album name)? And I have a work which continues onto disc 2 of a 2 disc release, so the heading appears twice. Can you do anything about that, or is the answer to tell people to tag the release as a single disc? (in the Qobuz plugin album browse, I ignore disc 1/disc 2 etc if we've got works).

darrell-k:
Thinking about this with respect to classical works like operas and ballets, where there's a 3 level hierarchy: 

Work -- Act -- track

instead of 

Work -- track

Post LMS9.0 we're going to attempt to handle this, but it's too late for the 9.0 release. For now there's a workaround:

The user can ignore the physical discs and instead number each act of the opera or scene of the ballet as a separate disc, and then use DISCSUBTITLE for the text.

So maybe not include the disc number ignore for now? We can tell people who hit the split work issue I described to retag the album as all one disc.

darrell-k:
Or even better, keep https://github.com/CDrummond/lms-material/commit/fb461ff397f567d19415d116f519c4066dcae31a but also implement a split in the play queue and tracks browse on the GROUPING tag which I'm pretty sure we're already providing in the tracks command response. (As I said, implementing this for the Default Skin UI is going to have to wait).

CDrummond:
Sorry - but what? Split on work_id, performance, **and** grouping? 3 extra tags just for Classical stuff????

darrell-k:
Well, yes. Classical music is inherently more complicated than album/track (especially when most classical discs contain multiple works and career-spanning box sets for well-known artists will only get more common as the record companies keep repackaging their ever-growing back catalogues...)  

I'm pretty sure that when LMS9 is released, though, we'll slowly build up an enthusiastic user base amongst classical collectors, because although I say it myself, we've created something really good here.

The Grouping thing could also be used for other genres: Side1/Side2 or Original Album/Bonus Tracks, etc.

CDrummond:
And as stated, I have no such music, hence why this is getting tedious - too many changes for an area I have no need for.

Anyhow, I've added splitting on grouping. So, is this done now?

darrell-k:
[...]

PR #993, opened 2024-11-29
Tweaks for Work/Grouping in play queue

darrell-k:
I wanted to push the Artist/Album text in the heading onto a second line, but couldn't work out how. 

Examples:

image

image

image

### DIFF
--- a/MaterialSkin/HTML/material/html/js/queue-page.js
+++ b/MaterialSkin/HTML/material/html/js/queue-page.js
@@ -40,6 +40,8 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
     let artistAlbum = undefined;
     let artistAlbumContext = undefined;
     let artistIsRemoteTitle = false;
+    let workGroupStr = i.work && i.composer ? i.composer+SEPARATOR+i.work : '';
+    workGroupStr += i.work && i.grouping ? SEPARATOR + i.grouping : i.grouping ? i.grouping : '';
     let artistStr = i.albumartist ? i.albumartist : i.artist ? i.artist : i.trackartist;
     if (queueAlbumStyle) {
         let id = i.albumartist ? i.albumartist_id : i.artist_id ? i.artist_id : i.trackartist_id;
@@ -69,8 +71,10 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
         artistAlbumContext = undefined;
     }
     if (!queueAlbumStyle || !artistIsRemoteTitle) {
-        artistAlbum = addPart(artistAlbum, buildWorkLine(i, artistStr, 'queue'));
         artistAlbum = addPart(artistAlbum, buildAlbumLine(i, 'queue'));
+        if (workGroupStr) {
+            artistAlbum = '<obj>' + workGroupStr + '</obj>'+ SEPARATOR + artistAlbum;
+        }
         if (queueContext) {
             let al = i18n('<obj>from</obj> %1', buildAlbumLine(i, "queue")).replaceAll("<obj>", "<obj class=\"ext-details\">");
             artistAlbumContext = undefined == artistAlbumContext ? al : (artistAlbumContext + " " + al);


MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -69,8 +71,10 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
         artistAlbumContext = undefined;
     }
     if (!queueAlbumStyle || !artistIsRemoteTitle) {
-        artistAlbum = addPart(artistAlbum, buildWorkLine(i, artistStr, 'queue'));
```
CDrummond, line 72: You have removed the call to `buildWorkLine` (bad name, as its not a line, but...). This is used to make the work part of the string clickable links (if enabled). This function is in `itemlinks.js`. For consistency you need to use this function, but feel free to adjust it as required (its only used here).

darrell-k: Ah yes, I stopped using that because I couldn't work out how to use `addPart` to put the "added part" at the beginning of the resulting string. I meant to go back to that once I got things looking the way I wanted but forgot.

What do you think about a 2 line heading in certain circumstances? Ideally only when dealing with a Work, not when simply a Grouping (like the Stranglers example)? Is it even possible? (I did try putting a `<br />` in the string, but that resulted in the first detail line being rendered on top of the second heading line.

If the 2 line option for Works is feasible, that would allow room for the standard Artist, Band, Conductor combination instead of just artist for classical works.

CDrummond: Not keen on a second line - the whole idea of this 'view' was to be compact. I'm not sure why you insist on having so much info on display. The first screenshot, to my mind, has way too much text. But that's just my opinion.

'addPart' always appends - this function is in `utils.js` Just changing the order should suffice, e.g. (not tested):

```javascript
artistAlbum = addPart(buildWorkLine(i, artistStr, 'queue'), artistAlbum);
```

darrell-k: The reason is that it is `Composer * Work * Grouping` which describes the group. 

I agree the first example looks messy, and it's not helped by there being only one track in the first group, which would almost always be the case for operas as they start with an overture. 2 lines might actually help here, obviously we don't want to lose the album link (though we could possibly move artist into the detail line - the one used in the header can be a bit random - changes with rescans - when there's more than one) and include in the detail line.

I would envisage the 2 lines only kicking in if we've got a work _and_ a grouping, so it's under the control of the user's tagging: my thinking is that if they've gone to the trouble of adding a Grouping tag, it's because they want to see it.

I'm guessing that to implement this the template would need to change around here:
https://github.com/CDrummond/lms-material/blob/e2bf13d5beddf50b7e59144a182df6dc19addb41/MaterialSkin/HTML/material/html/js/queue-page.js#L253

CDrummond: For 2 lines you also need to set the size of each item - see the "size:" setting on line 169

However, if multi-lines are required then perhaps it's just not suitable to this view style?

darrell-k: I think for "tag heavy" stuff like operas, the hierarchical view is better suited than repeating everything on every track line. But let me throw something together and then you can see what you think.

darrell-k: I've opened up space for the track title:

image

but have spent hours messing with the template but can only seem to push the text down into the gap by changing the stylesheet (and the track time does not go with it):

image

What am I missing?

darrell-k: > Also, in the above, why repeat "Violin Concerto in D Minor, Op. 47" on each line?

That's my track name tag. In my test database I have a mixture of it set to WORK:MOVEMENT  and just MOVEMENT, so I can test the variations that I know people use out there.

CDrummond: Sorry, that I'm not more helpful at the moment - I'm busy at work and at home at the moment! If you work on splitting the text into what's required I'll look at the CSS / template changes.

darrell-k:
We should probably put this on hold for now (and also back out the commits you made for #992) because the whole work/grouping thing is causing some problems for MusicBrainz/Picard users who have those tags and didn't necessarily know it, until LMS9 started using them.

CDrummond:
Why? What's wrong with the current code?

darrell-k:
It's grouping the play queue by work (if that exists). This could potentially cause problems for releases tagged by Picard: unless you're using one of Picard's classical music plugins it will generally get WORK wrong, and for non-classical music, WORK has a different meaning in MB world. 

The assumption that I was working on until now, that if the user has WORK/GROUPING tags it's because they wanted them, has turned out to be faulty: some Picard/MB users are discovering for the first time that their collection has these tags.

It's not a problem with browsing works, because the user can switch off Works in the interface prefs.

CDrummond:
Changes reverted. Or rather, there still there just commented out...

PR #997, opened 2024-12-04
Handle multiple genres in play queue

darrell-k:
I'm still playing with https://github.com/CDrummond/lms-material/pull/993 but along the way I found something to fix: if a track is tagged with multiple genres, the value returned in `genre` will be randomly one of them. So get and use `genres` instead.

### DIFF
--- a/MaterialSkin/HTML/material/html/js/utils.js
+++ b/MaterialSkin/HTML/material/html/js/utils.js
@@ -1041,7 +1041,8 @@ function getClickPos(ev) {
 }
 
 function useArtistTagType(genre, genres) {
-    return (genre && genres.has(genre)) || (1==genres.size && genres.has('*'));
+    const genreArray = genre.toString().split(/\s*,\s*/);
+    return (genreArray.some(g => genres.has(g))) || (1==genres.size && genres.has('*'));
 }
 
 function useComposer(genre) {
--- a/MaterialSkin/HTML/material/html/js/queue-page.js
+++ b/MaterialSkin/HTML/material/html/js/queue-page.js
@@ -6,7 +6,7 @@
  */
 'use strict';
 
-const PQ_STATUS_TAGS = "tags:cdegilqtuy" + (LMS_VERSION>=90000 ? "bhz1" : "") + "AAIKNSxx";
+const PQ_STATUS_TAGS = "tags:cdegilqtuy" + (LMS_VERSION>=90000 ? "bhz1" : "") + "AAIKNSxxGP";
 const PQ_REQUIRE_AT_LEAST_1_ITEM = new Set([PQ_SAVE_ACTION, PQ_MOVE_QUEUE_ACTION, PQ_SCROLL_ACTION, PQ_SORT_ACTION, REMOVE_DUPES_ACTION]);
 const PQ_REQUIRE_MULTIPLE_ITEMS = new Set([PQ_SCROLL_ACTION, SEARCH_LIST_ACTION, PQ_SORT_ACTION, REMOVE_DUPES_ACTION]);
 
@@ -52,9 +52,9 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
             artistAlbum = buildLink(i.albumartist ? 'show_albumartist' : 'show_artist', id, artistStr, 'queue');
         }
     } else {
-        let useComposerTag = i.composer && lmsOptions.showComposer && useComposer(i.genre);
-        let useConductorTag = i.conductor && lmsOptions.showConductor && useConductor(i.genre);
-        let useBandTag = i.band && lmsOptions.showBand && useBand(i.genre);
+        let useComposerTag = i.composer && lmsOptions.showComposer && useComposer(i.genres);
+        let useConductorTag = i.conductor && lmsOptions.showConductor && useConductor(i.genres);
+        let useBandTag = i.band && lmsOptions.showBand && useBand(i.genres);
         artistAlbum = buildArtistLine(i, 'queue', undefined, undefined, useBandTag, useComposerTag, useConductorTag);
         artistAlbumContext = queueContext ? replaceBr(buildArtistWithContext(i, 'queue', useBandTag, useComposerTag, useConductorTag), " ") : undefined;
     }


darrell-k:
Some more places to look at:
```

browse-resp.js:1400:                let useComposerTag = i.composer && lmsOptions.showComposer && useComposer(i.genre);
browse-resp.js:1784:                let useComposerTag = i.composer && lmsOptions.showComposer && useComposer(i.genre);
nowplaying-functions.js:173:    let useComposerTag = playerStatus.current.composer && lmsOptions.showComposer && useComposer(playerStatus.current.genre);
nowplaying-functions.js:255:    let maiComposer = lmsOptions.maiComposer && view.playerStatus.current.composer && undefined!=view.playerStatus.current.composer_id && useComposer(view.playerStatus.current.genre);
nowplaying-functions.js:350:        if (lmsOptions.showComposer && view.playerStatus.current.composer && view.playerStatus.current.composer_id && useComposer(view.playerStatus.current.genre)) {
queue-page.js:58:        let useComposerTag = i.composer && lmsOptions.showComposer && useComposer(i.genres);
itemlinks.js:136:    if (undefined!=useComposerTag ? useComposerTag : (i.composer && lmsOptions.showComposer && useComposer(i.genre))) {

```

CDrummond:
Perhaps the real fix is for LMS to return the first genre when requested, not just picking a random one. That at least would make sense. Not sure I really want to be checking a set of genres for every track in the queue - there could be thousands!

darrell-k:
I think thousands is unlikely, but there could be a few. (Or do you mean thousands of tracks, rather than genres?)

The behaviour of LMS is a "feature" of sqlite: unlike some databases it allows selection of a column that's not in the "GROUP BY" but the value that column gets is "undefined" according to the docs. This is noted as a bug in the LMS source, and I think at a later stage a subquery was added to populate genres with the full list.

If it was me, I'd have removed the old single genre at that point, but backwards compatibility, I suppose.

Technically, it would be easy enough to replace the random value of the returned genre with a chosen single value, but the problem is, which one to choose? We don't have a hierarchy of genres defined anywhere, and I don't think it's something that could ever be agreed upon to everyone's satisfaction.

Realistically, how many genres is a track going to have? 3? 4? I think more than that would be pretty rare. So IMO the extra processing overhead would not be noticeable.

The alternative would be for the user to add all their more specific genres to the lists in Material's server-side prefs, I see that you have quite a few already there by default (but sods law, not the one I was using). But I'm always in favour of things "just working" without the user having to think too much and manipulate prefs, as far as possible.

CDrummond:
Yes I meant thousands of tracks. 

The point still stands that if LMS has a flag to return a single genre it should be consistent. Either the first, or the more prominent (however that is decided). Just returning any random genre sounds like a bug to me.

darrell-k:
As I said, it is a bug, noted as one in the source. I think the fix was adding the subquery to return all the tracks genres. I assume the old code is still there for backward compatibility with hardware squeeze devices for example.

MaterialSkin/HTML/material/html/js/utils.js
```diff
@@ -1041,7 +1041,8 @@ function getClickPos(ev) {
 }
 
 function useArtistTagType(genre, genres) {
-    return (genre && genres.has(genre)) || (1==genres.size && genres.has('*'));
+    const genreArray = genre.toString().split(/\s*,\s*/);
```
CDrummond, line 1044: Don't like this converting to array each and every time a tag type is checked. You should change the `splitMultiples(i);` call @line 103 in `queue-page.js` to ` splitMultiples(i, true);` - this _should_ then convert `genres` into an array. This `useArtistTagType` could then be:

```javascript
function useArtistTagType(genres, tagGenres) {
        return (1==tagGenres.size && tagGenres.has('*')) || (genres.some(g => tagGenres.has(g)));
}
```

CDrummond:
I've created branch `997-multiple-genres` which should implement these changes. Can you checkout and confirm this is OK? Thanks.

CDrummond:
Changes form my branch merged into master, so closing.

darrell-k:
Sorry, missed your previous message, and haven't tried it out yet, but does passing `i` instead of `i.genres` to the utils function get around what I found, that `i.genres` could sometimes be an object, which is why I had to do the `toString`?

CDrummond:
Not sure how it could be an object? Never seen that. But my code to split the field does cast to a string first:

```javascript
function splitStringArray(val, plainSep) {
    return undefined==val || Array.isArray(val) ? val : (""+val).split(plainSep ? "," : MULTI_SPLIT_REGEX).map(function(itm) { return itm.trim() });
}
```

...this is used split all multi-string fields. (If only LMS would return an array!  (Which the above code should handle)).

Now I'm wondering if i should change `""+val` to `val.toString()` ???

darrell-k:
I recreated it:

with:
`const genreArray = genre.split(/\s*,\s*/);`

everything seems ok in the UI, but you get this repeatedly while a track is playing:

```
vue.min.js?r=2.7.15:11 TypeError: genre.split is not a function
    at useArtistTagType (utils.js?r=GIT-ccc790a0a:1044:30)
    at useComposer (utils.js?r=GIT-ccc790a0a:1049:12)
    at buildArtistAlbumLines (queue-page.js?r=GIT-ccc790a0a:58:71)
    at a.<anonymous> (queue-page.js?r=GIT-ccc790a0a:530:82)
    at pn (vue.min.js?r=2.7.15:11:22017)
    at t.$emit (vue.min.js?r=2.7.15:11:44399)
    at a.handlePlayerStatus (server.js?r=GIT-ccc790a0a:629:17)
    at server.js?r=GIT-ccc790a0a:880:26
```

CDrummond:
Is that error with the current code? Or before? As I use `(""+genre).split(...`

darrell-k:
That is with my PR (this one), but taking out the `toString` in my version of `useArtistTagType`. I'll have to pull your latest changes and try it out.

BTW, we tried returning an array from LMS in https://github.com/LMS-Community/slimserver/pull/1195 but it didn't get past Michael!

darrell-k:
All seems OK with your latest master. No console errors.

PR #1001, opened 2024-12-11
subheading, work,grouping,performance WIP

darrell-k:
Transferred from #993 

I hope I've got it right, my local git was in a bit of a mess, as usual.

I've attached a test album which has 3 works on 2 discs, the first work runs over onto the second disc. You can test the grouping stuff with any old album tagged with Grouping set to eg "Side A" or "Side B".

Three issues I know about:

1. As mentioned before, needs template/css mods when the albumStyle header goes onto 2 lines.
2. Play Queue operatons from the work or grouping do not work (see JS console output below).
3.  The popup when adding the whole album to the play queue from the album list still has the option "All Discs" rather than "All Works" or "All Groups".

And the coding style is quite bad!
[Alicia.zip](https://github.com/user-attachments/files/18099978/Alicia.zip)

```
vue.min.js?r=2.7.15:11 TypeError: Cannot read properties of undefined (reading 'id')
    at browseDoListAction (browse-functions.js?r=5.6.0:2650:17)
    at browseDoList (browse-functions.js?r=5.6.0:2639:9)
    at a.doList (browse-page.js?r=5.6.0:1601:13)
    at browseItemAction (browse-functions.js?r=5.6.0:1591:18)
    at a.itemAction (browse-page.js?r=5.6.0:1059:17)
    at click (eval at cc (vue.min.js?r=2.7.15:1:1), <anonymous>:3:32217)
    at pn (vue.min.js?r=2.7.15:11:22017)
    at HTMLImageElement.r (vue.min.js?r=2.7.15:11:10396)
    at i._wrapper (vue.min.js?r=2.7.15:11:59549)
```

### DIFF
--- a/MaterialSkin/HTML/material/html/js/browse-resp.js
+++ b/MaterialSkin/HTML/material/html/js/browse-resp.js
@@ -1412,7 +1412,25 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                         subtitleContext=i18n('<obj>from</obj> %1', buildAlbumLine(i, "browse", false)).replaceAll("<obj>", "<obj class=\"ext-details\">");
                     }
                 }
-                if (undefined!=i.disc && !isSearchResult && !isAllTracks && !isCompositions) {
+                if ( (undefined!=i.work && undefined!=i.composer || undefined!=i.grouping) && !isSearchResult && !isAllTracks && !isCompositions) {
+                    let discNum = undefined;
+                    if (i.composer && i.work) {
+                        discNum = i.composer+SEPARATOR+i.work;
+                        if (i.performance) {discNum += SEPARATOR+i.performance};
+                        if (i.grouping) {discNum += SEPARATOR+i.grouping};
+                    } else {
+                    	discNum = i.grouping;
+                    }
+                    if (discs.has(discNum)) {
+                        var entry = discs.get(discNum);
+                        entry.total++;
+                        entry.duration+=duration;
+                    } else {
+                        let title = discNum;
+                        discs.set(discNum, {pos: resp.items.length, total:1, duration:duration, title:title});
+                    }
+                }
+                else if (undefined!=i.disc && !isSearchResult && !isAllTracks && !isCompositions) {
                     let discNum = parseInt(i.disc);
                     if (discs.has(discNum)) {
                         var entry = discs.get(discNum);
--- a/MaterialSkin/HTML/material/html/js/itemlinks.js
+++ b/MaterialSkin/HTML/material/html/js/itemlinks.js
@@ -238,16 +238,22 @@ function buildAlbumLine(i, page, plain) {
     }
 }
 
-/*
-function buildWorkLine(i, artist, page, plain) {
+
+function buildWorkLine(i, page, plain) {
     var line = undefined;
-    if (i.work && artist) {
-        var work = i.work;
-        if (i.work_id && (!IS_MOBILE || lmsOptions.touchLinks) && !plain) {
-            work="<obj class=\"link-item\" onclick=\"showWork(event, "+i.work_id+",\'"+escape(work)+"\',\'"+escape(artist)+"\', \'"+page+"\')\">" + work + "</obj>";
-        }
+    if ((!IS_MOBILE || lmsOptions.touchLinks) && !plain && i.work && i.composer) {
+        var work="<obj class=\"link-item\" onclick=\"showWork(event, "+i.work_id+",\'"+escape(i.work)+"\',\'"+escape(i.composer)+"\', \'"+page+"\')\">" + i.composer + SEPARATOR + i.work + "</obj>";
         line=addPart(line, work);
     }
     return line;
 }
-*/
+
+function buildGroupingLine(i, page, plain) {
+    var line = undefined;
+    if ((!IS_MOBILE || lmsOptions.touchLinks) && !plain && i.grouping) {
+        var grouping = "<obj>" + i.grouping + "</obj>";
+        line=addPart(line, grouping);
+    }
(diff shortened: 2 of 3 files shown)


CDrummond:
Thanks for this, but if you don't mind I think I'll use these changes to re-implement this myself. I think if an album is split across discs then that splitting has to take precedence - as when just listing albums Material will show the multi-disc icon, and pressing add/play will prompt for which disc. To do likewise for works would mean needing a genre for the album - to know that it is classical with works, which (AFAIK) is not currently possible.

CDrummond:
And I see you example is multi-disc. Sorry, but I cannot see splitting on disc and works....

darrell-k:
Absolutely no problem with you re-implementing, you'll do a much better job, technically.

And yes, splitting by disc and work etc at the same time is not feasible - in the Qobuz plugin, it's either one or the other.

But I think the work/grouping/performance split should take precedence: if the user has gone to the trouble of tagging work/grouping/performance, it's surely because that's the way they see their collection?

Splitting on work only for single disc releases would lead to strange results - why would one release group by work, but others not?  There are many, many multi-disc albums in the classical world, and just about any opera (which is a very important use of grouping) will be multi-disc, very few operas fit on a single CD.

CDrummond:
Initial track grouping done.

As stated grouping by disc **has** to take precedence, as its what's used at the upper level. If not then when listing albums you'd see the multi-disc icon and be prompted for which disc to play, but then when you navigate in the split has now changed to works??? That would be an awful UI. I get that works probably span discs - but in that case what is the real use case of preserving the by disc split? Why not just re-tag to single disc?

darrell-k:
Well, the multi-disc icon is neither here nor there, it could just mean a long release. And surely it could prompt for which work/grouping to play it that's in the tags?

As I said above, going disc first will lead to massive inconsistencies.  Yes, retagging as a single disc would be a workaround, but I'm trying to avoid an almost mandatory retagging step for users (for whatever reason, even when you buy downloads, they're often tagged with disc number).

CDrummond:
Again, no - because at the album level I only know if an album is multi disc not if it is a multi work classical album. So disc has to take precedence for consistency.

darrell-k:
But this is what I get with my implementation: the only problem seems to be the text of the first entry, which says "All Discs"???

(This release has 3 discs, 1 work, but I've used a GROUPING tag to split into the overture and 2 acts (which don't correspond to the disc boundaries)

image

CDrummond:
But you only get that because it's also multi disc. If you had a multi work album on a single disc no prompt will show. That's the issue.

CDrummond:
If LMS could return a (e.g) "workcount" field (similar to "disccount") and "isClassical" in an "albums" response then this could be as you wish. But not as it current stands.

darrell-k:
Well, I'm sure that could be arranged.  maybe a single "groupcount" which takes account of work/grouping/performance, or even just a hasGroups flag? Do you need the actual number?

On the isClassical thing, it's problematic at album level, as multi-genre releases cannot be ruled out. Something like a Billie Holiday box set could contain Jazz, Classical and Blues, for example. 

Now that we're controlling the scanning of works based on genre, I don't think you need to check the genre at this level anyway: There won't be works unless the user has specified that they want them.

Also, there's the question of GROUPING for rock/pop albums - Bright Side/Dark Side on Peter Gabriel's latest release for example, or to split the bonus tracks or live tracks you often get on a CD reissue from the main album. This why it was `work && composer || grouping` in my implementation (I think your code is slightly wrong there, its work and composer that always go together).

So I don't think isClassical should be in there.

CDrummond:
No, no actual count is needed - just an indication that it is multi whatever. The whatever, however, is useful so that I can say "All xxx" and show a relevant multi icon. So, for now mult-works is enough.

I always get my operator precedence wrong, hence using brackets to make it explicit. I'll update the code, along with some other fixes.

darrell-k:
I've created PR https://github.com/LMS-Community/slimserver/pull/1249

Available now at https://github.com/darrell-k/slimserver/tree/group-count

Pass tag `2` to get group_count, which is a count of distinct work/group/performance for the album.

CDrummond:
Can you test the `dev` branch. Is this sufficient? It also adds grouping in the queue for non-works. Not 100% sure about that, but added for consistency with browse.

darrell-k:
OK, I'll give it a go right now. I've just pushed the amendments for https://github.com/LMS-Community/slimserver/pull/1249

darrell-k:
Something wrong with the "standard" play queue display:
image

CDrummond:
Hmm!!! Will fix. But, what about album-style queue and browsing?

CDrummond:
Fixed now.

darrell-k:
I'm still going through various combinations of work/performance/grouping but It's looking very good. Thanks. People are going to be very happy with this.

One small point: should we lose the multi-disc icon when there's only one disc?

CDrummond:
I probably need a better icon. I created a multi-works icon, but as you noted this is not restricted to works. Just not sure what a "multi group" icon would look like - hence I left the disc icon. It needs an icon as the behaviour changes because of the grouping.

darrell-k:
Maybe just $$$ if you imagine the dollar sign to be a treble clef?

One thing, not really related to this change, when the colour scheme picked up from the coverart is monochrome (Black And White, for example!) the headings in the track listing don't really stand out:
image

CDrummond:
Problem is Material extensively uses treble-clef for classical items.

What about the icon now?

CDrummond:
Anyhow, I take it this and #993 + #992 can be closed?

darrell-k:
Yes, that icon is fine. And yes, close the PRs and the issue.

Thanks again for this Craig.

PR #1003, opened 2024-12-16
Adjust album cover size on NP screen for tablets

denys-lysiuk:
Hi Craig, I've tried to adjust album cover sizes in portrait view for desktop - looks like they can use a little bit more space. This seems to do the trick, not sure about the purpose of 50vh check though. Tried different tablet devices simulated in browser devtools - looks correct. E.g. iPad:

before:

ipad_before

after:

ipad_after


CDrummond:
Thanks for the pull, but not 100% sure about this - looks a little cramped. I have, however, increased the max size of portrait image.

PR #1004, opened 2024-12-16
Fix URL parsing

michaelherger:
Don't know whether you've copied that regex from my implementation in LMS. But it seems to be missing the `%` sign, too. This change should fix https://forums.slimdevices.com/forum/user-forums/3rd-party-software/1740806-advanced-tag-view-urls-not-always-shown-correctly#post1740984


CDrummond:
I think you must've given me that regex - as there is no way I wrote that!

PR #1008, opened 2024-12-19
Fix descending sort toggle

darrell-k:
I noticed a problem testing my sort enhancements thing, and I think it's an existing problem: when changing the toggle for descending, local storage was not being written properly, it ended up undefined. I _think_ this fixes it.


CDrummond:
Thanks!

Issue #1007, opened 2024-12-17
QUESTION works-album sort?


darrell-k:
A question, not an issue.

I'm working on an LMS enhancement to allow new sort options (by role) for the works-album list (eg see below in Home/My Music/Works/Composers/Bartok).

The idea is to call the new roles query to find out which roles are available for the list shown and populate the sort list with them.

Could also be used outside of works in an Artist/Releases list or even the main Releases list.

I'm trying to test it out before requesting a Material change, but how do I enable the sort option in the ... menu on this screen, as is available in the Artist/Albums view? 

![image](https://github.com/user-attachments/assets/cb5057e5-6eae-4480-94e4-a9fa766c713d)

CDrummond:
This is excluded by line 1094 in browse-page.js - if you comment out that line you will see the album-sorts action.

darrell-k:
Thanks. I just uncovered a bug, I think:

When changing the sort in an artist releases list, there is an error if there is a works section:

```
[Wed, 18 Dec 2024 11:07:20 GMT] ERROR:TypeError: Cannot read properties of undefined (reading 'id') (5) ['albums', 'tags:aajlqswyKRSSW2E', 'sort:album', 'artist_id:558487', 'library_id:-1']
logError @ utils.js?r=GIT-b659c0dc5:156
logAndShowError @ utils.js?r=GIT-b659c0dc5:161
(anonymous) @ browse-page.js?r=GIT-b659c0dc5:1347
Promise.catch
refreshList @ browse-page.js?r=GIT-b659c0dc5:1346
sortItems @ browse-page.js?r=GIT-b659c0dc5:1284
click @ VM109537:3
pn @ vue.min.js?r=2.7.15:11
r @ vue.min.js?r=2.7.15:11
pn @ vue.min.js?r=2.7.15:11
(anonymous) @ vue.min.js?r=2.7.15:11
click @ vuetify.min.js?r=1.5.24trim20241031:6
pn @ vue.min.js?r=2.7.15:11
r @ vue.min.js?r=2.7.15:11
(anonymous) @ vue.min.js?r=2.7.15:11
utils.js?r=GIT-b659c0dc5:157 console.trace
logError @ utils.js?r=GIT-b659c0dc5:157
logAndShowError @ utils.js?r=GIT-b659c0dc5:161
(anonymous) @ browse-page.js?r=GIT-b659c0dc5:1347
Promise.catch
[...]

PR #1011, opened 2024-12-21
Sorting by role

darrell-k:
PR for https://github.com/CDrummond/lms-material/issues/1007

Add role sort options for Albums and Album-works.

LMS 9.1 from https://github.com/darrell-k/slimserver/tree/sorting-by-role (just updated) is required.


MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -1882,6 +1882,39 @@ function browseHeaderAction(view, act, event, ignoreOpenMenus) {
         for (var i=0,len=sorts.length; i<len; ++i) {
             menuItems.push({key:sorts[i].key, label:sorts[i].label, selected:sort.by==sorts[i].key});
         }
+
+        if (LMS_VERSION>=90100 && ALBUM_SORTS_ACTION==act) {
```
CDrummond, line 1886: Calling this every time the menu shown is inefficient. The result of this call should be stored with the current item, and reused if possible.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1010,8 +1010,17 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                         }
                     }
                 }
+                // sort field for role sorting might be null, if so make sure nulls stay below non-nulls even when reversed.
+                let loopNotNulls = [];
+                let loopNulls = [];
+                const firstNullPos = data.result.albums_loop.map(e => e.rolesort_name).indexOf('');
+                console.log("DK albums_loop firstNullPos="+firstNullPos);
```
CDrummond, line 1017: Remove.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1010,8 +1010,17 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                         }
                     }
                 }
+                // sort field for role sorting might be null, if so make sure nulls stay below non-nulls even when reversed.
+                let loopNotNulls = [];
```
CDrummond, line 1014: This is only of use if reversed (and not new music), no?

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1010,8 +1010,17 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
...
+                let loopNulls = [];
+                const firstNullPos = data.result.albums_loop.map(e => e.rolesort_name).indexOf('');
```
CDrummond, line 1016: Out of curiosity, what impact does this lookup have on loading large lists?

MaterialSkin/HTML/material/html/js/browse-functions.js
darrell-k: I was trying doing this, but as soon as I started pushing to an array other than `menuItems`, the application would no longer wait for the server command results. The problem is I don't understand how the async stuff works.

MaterialSkin/HTML/material/html/js/browse-resp.js
darrell-k: It's bound to slow it down, I'll test with a 25,000 size albums list...

MaterialSkin/HTML/material/html/js/browse-functions.js
CDrummond: It does not 'wait', the menuItems is 'reactive' - so vue.js will update UI when anything changes. What you need to do is copy the (parsed) details of this response somewhere, and **also** update menuIitems. If this cached version exists, then use that for menuItems.

MaterialSkin/HTML/material/html/js/browse-resp.js
CDrummond: Well then it depends on how much - as this is not something I'd ever really use. However, I really cannot see if having a significant impact. As stated, I was just curious.

MaterialSkin/HTML/material/html/js/browse-functions.js
darrell-k: Ah, I understand, it's in the framework. No wonder I couldn't find it!

MaterialSkin/HTML/material/html/js/browse-resp.js
darrell-k: You're right, negligible: <1ms on a mid-power laptop. Rendering 25,000 items takes much longer. I suppose it's doing a binary chop on the array.

MaterialSkin/HTML/material/html/js/browse-functions.js
darrell-k: OK, I'm about to push some changes which hopefully cache the roles command results correctly.

These changes require the LMS from my updated branch https://github.com/darrell-k/slimserver/tree/sorting-by-role

I've also decided to clear any role sort when changing the view, writing the default to local storage for new views (except if we're viewing all albums, because by definition all roles sorts would be available).

This avoids having a stored sort (in local storage) which might not apply to a new view, because the set of albums do not contain the previously-selected sort role.

There is commented out code for a different approach, which "pretends" the sort is the default 'album' one if the actually used sort is on a role which isn't used in a new album list. This works because the LMS query will return the albums_loop in album name sequence for a role sort where the role doesn't exist for any albums. This means if the user subsequently displays an album list where the previously used sort role exists, it would still be used. But this is a bit messy and possibly fragile.

I suppose the ideal solution would be to have `getAlbumSort` always validate a role sort against the current albums list, but I think that would mean really having to wait for the LMS rolesQuery to return.

Don't know what you think on this?

Anyway, lots more testing to do, but I thought I'd push this so you can see where I am right now. (And I know I have a few debug messages to remove!)

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -2540,7 +2598,8 @@ function browseReplaceCommandTerms(view, cmd, item) {
                 }
             } else if (cmd.params[i].startsWith(SORT_KEY+ALBUM_SORT_PLACEHOLDER) ||
                        cmd.params[i].startsWith(SORT_KEY+ARTIST_ALBUM_SORT_PLACEHOLDER)) {
-                var sort=getAlbumSort(cmd, view.inGenre);
+console.log("DK item.id="+item.id);
+                var sort=getAlbumSort(cmd, view.inGenre, undefined, item.id=='mm:/myMusicAlbums');
```
CDrummond, line 2602: Why is "myMusicAlbums' getting sorts for all roles? I'm not sure it makes sense to sort the LMS albums lit by (e.g.) 'Guitarist' - that info is never shown.

darrell-k: If you choose a role sort, LMS will return the sorted role person as the first artist in `artists` and `artist_ids`, causing the name to show first in the album list artist links.

CDrummond: Still, sounds like a change just because it can be done - not one, IMVHO, that make sense. I see a list of albums - ordering by _Album_Artist, Release date, or title make sense, anything else not so much. Why, with this particular list, would I want to sort by (e.g.) Guitarist? Makes no sense.

darrell-k: Yeah, I did mention earlier in this thread that I wasn't sure about role sorts at the top level albums list. But once I removed artist/albumartist/trackartist from the generated sorts, it looked OK to me, so I thought "why not?". It's not inconceivable that someone might want to see a list of their albums by guitarist, producer or whatever.

CDrummond: Sorry, not convinced. Just because you can do something does not mean you should.

darrell-k: I just think that if a user has gone to the trouble of creating specific role tags, it's because they're important to them (and would not likely leave them unused). But this doesn't apply to the standard roles (composer/band/conductor), they're there anyway. I'll make some changes.

CDrummond: Still does not mean it makes sense for this list.

darrell-k: Looking back at this, All that code is doing is ensuring that a previously-selected album sort is held when displaying all albums, on the basis that it has to be valid - see comment above:

> I've also decided to clear any role sort when changing the view, writing the default to local storage for new views (**except if we're viewing all albums, because by definition all roles sorts would be available**)...

I suppose the wider point you are making is that you don't think the role sort options should be available in the all albums list at all. But why not there if we offer them (for example) when browsing albums for a particular genre? In the end, it's just some extra options added to the drop-down sort menu.

CDrummond: Because this very much sounds like adding something just because you can. In reality when would you ever want to order all your albums by the Guitarist? Why would you do that as opposed to entering the "Guitarist" browse mode?

...and the same goes for "Albums" under "Genres".

Do any other music servers add such a feature? As its not something I can personally see the point to. And I don't want to overcrowd a menu just because we can.

darrell-k: > Why would you do that as opposed to entering the "Guitarist" browse mode? ...and the same goes for "Albums" under "Genres".
Because it would avoid having to set up Additional Browse Menus for each role by Album (in addition to Role by Artist), and allow the user to quickly find different roles in whatever set of albums they are currently viewing rather than having to go back to the main menu and select a new browse mode each time. And even after doing that, the album list would not be sorted by the browsed role, only restricted to it.

CDrummond: I'm probably being slow here, so sorry for being a pain, but... and? **Why** would you want to sort albums by Guitarist? Artist, yes, but Guitarist? Makes no sense. Why would you want role-by-album??? If you care about Guitarist then you'd want Guitarist -> ABC -> albums. But sorting ALL albums by Guitarist? No, don't see the use-case. this very much sounds like "we have all these new role options, lets put these in all the possible places they _could_ (not _should_) go"

(I'm using Guitarist as an example, I realise there are many more roles)

Again, does **any** other media server do such a thing (by default)?

darrell-k: On other media servers I don't know, I don't use anything but LMS. A user did mention one which was more flexible than LMS, in that it allowed browsing by Album then Artist Role (or any arbitrary hierarchy - I think it might have been Minim), as well as Artist Role then Album, which is sort of in the same area.

Now, LMS does not offer this directly, but through the sort mechanism, we can offer the route.

For example, if I wanted to see all my jazz clarinet stuff, I could go to Genres -> Jazz -> Releases and then sort by Clarinet Player. Of course there are other routes, but I think this is a valid and useful one.

Bear in mind that only roles used on the albums being displayed will be added to the sort options. So if the user hasn't tagged composer/conductor/band or created user-defined roles for anything in the list, they won't see anything other than the existing sort options. And if they have tagged those roles it likely they are important to them. So it's not going to bombard the user with sort options irrelevant to them.

The origin of this PR was a request by users with a lot of instances of a work, and after displaying the works-album list they want to be able to sort the performances by orchestra, conductor, soloist, whatever. We could restrict the additional sort options to work-albums only, but as you will have worked out by now, I think that would be a shame, and once it's there for work-albums someone is bound to ask why it's not there for albums in general.

Another way of looking at this:

LMS long ago went beyond just Album/Artist/Year, with the introduction of Composer, Conductor and Band, and now we have user-defined roles. For many users, the days when Artist is the only (or even most important) role in a library are long gone.

But LMS sort options didn't keep up: the list of available sorts was hardcoded in the albums query, and it would throw an error if it got a sort parameter that it didn't recognise. On that basis, Material has the same hardcoded list. But now, if LMS is to allow sorting by these additional roles, why would Material not?

CDrummond:
How close to completion is this?

Not sure this is possible, and not worth recreating the pull request if not, but can the target branch be changed to dev? I've just created a new 'dev' branch for 5.8.0

darrell-k:
I'm currently making the changes on the Default Skin side to allow role sorting options. How close do I need to be?

CDrummond:
Not sure I get what you mean by "How close do I need to be?" ?? I was just wondering if this pull request is complete or not.

I make releases from the master branch, so don't want to merge this feature change into there just yet. I usually make feature changes in a dev branch, and only when ready to I merge that into master. Hence my asking if it'd be possible to change the target branch, but I'm not sure if this is possible with github...

darrell-k:
OK, I can try to change the target, I'm pretty sure I've done it before.

CDrummond:
Any updates with regards to this? Is it finished? Does it require a corresponding LMS change?

I've also merged `dev` into `master`, so no need to change target branch (unless I release 5.8.0 before this is merged).

darrell-k:
I've got diverted by bug fixes etc, but I will continue with this (and rebase as necessary) asap. I want to implement the same thing for Default Skin, and those changes may simplify this.

CDrummond:
I'm going to close this. It's been inactive for over a year, and is not a feature I can personally see the use-case for.

PR #1025, opened 2025-01-25
limit works advanced search to discovered albums

darrell-k:
return album ids in advanced search works_loop and use them to limit the instances of works returned to the user. Requires https://github.com/LMS-Community/slimserver/pull/1309

### DIFF
--- a/MaterialSkin/HTML/material/html/js/advancedsearch-dialog.js
+++ b/MaterialSkin/HTML/material/html/js/advancedsearch-dialog.js
@@ -415,6 +415,7 @@ Vue.component('lms-advancedsearch-dialog', {
                 let workPos = -1;
                 let results = [];
                 let total = 0;
+                let albumIds = [];
 
                 // Rename loops so that parseBrowseResp handles in
                 // requested order...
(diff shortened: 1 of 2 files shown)


CDrummond:
Is this OK to apply now? As in does search still work as before, just this improves things? Or does it need an LMS version check?

darrell-k:
Should be OK, but just pushed a change on the Perl side to check we've got the data.

darrell-k:
Actually it's still producing an error in the LMS log, hold on for now...

darrell-k:
Now it checks properly!

darrell-k:
I'm having a bad afternoon, now it's not passing album_id at all!

darrell-k:
Added version check in Plugin.pm. (`has_column` is always false for a derived column as `albumId` is, so can't use that). It will log an error until 9.0.2 users update their LMS, but the old functionality still works.

darrell-k:
@CDrummond any thoughts on the latest commit here?

CDrummond:
@darrell-k I was waiting for the feature to be merged into LMS. Everything looks OK to me.

darrell-k:
Thanks. LMS change should be in tonight's 9.02.

CDrummond:
Merged then :)

PR #1049, opened 2025-03-06
Add support for `portraitid` in LMS 9.1

michaelherger:
~~@CDrummond this by far is not ready to be merged. But it's my ongoing WIP on artist picture support. Please feel free to comment if that's totally off.~~

@CDrummond I've merged the artist portrait code to LMS 9.1. It should fix the issue we've discussed in the past where artist IDs were re-used and thus invalid pictures were shown. 9.1 uses code borrowed from the cover art handling, where the image's timestamp etc. are taken in to account to create a unique ID for the pictures.

Please note that while this should be backwards compatible with LMS9.x, it requires ~~my still not merged https://github.com/LMS-Community/slimserver/tree/artist-artwork branch~~ the latest LMS 9.1 to take advantage of the LMS own artist picture support.


michaelherger:
I've had this in place for a week or two. Should work just fine.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1208,6 +1211,8 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
             showRoles.delete(TRACK_ARTIST_ROLE);
             ignoreRoles.forEach(role => { showRoles.delete(role) });
             resp.showRoles = Array.from(showRoles).sort();
+            // TODO - don't know how to handle a portraitId here. We can have an album's main artist picture. But that's not always suitable here.
+            // Using the artist ID could cause caching issues again.
```
michaelherger, line 1215: @CDrummond do you have an idea how we could address this?

CDrummond: Can `portraitid` not be returned in the response? e.g.

```json
{
    ...
    "portraitid":12345,
    "albums_loop": {
     ....
    }
    ...
}

Comment (CDrummond) [reply MaterialSkin/HTML/material/html/js/browse-resp.js:1215]: ...and if more than 1 artist id, then return them all? e.g.

```json
{
    ...
    "portraitids":[12345,2456],
    "albums_loop": {
     ....
    }
    ...
}
```

perhaps, for simplicity, always return a list for `portraitids` even if there is only one?

michaelherger: I don't remember in what exact context we would hit that code. I thought it wasn't an artist query, but something else? Isn't this being used when showing the biography or something?

CDrummond: Its not used much - AFAIR, only when showing a list of an artist's albums when 'go to artist' is used. So, the user has _not_ browsed "My Music / Artists / TheArtist" and therefore there is no known artist image to use, so I use the artist id of the request to get the image (which is shown in the browse toolbar).

CDrummond: @michaelherger Scratch that. For these "go to artist" calls I could first call "['artists', 0, 1, 'tags:xxx', 'artist_id:xxx']" to get portraitid - then make the albums loop call. I'll update Material for this - I've raised #1054 for this.

PR #1050, opened 2025-03-07
Simplify player volume control determination

SamInPgh:
Replace the 'has_digital_out' status query attribute with 'use_volume_control', a more straightforward approach that places the responsibility for determining a player's volume behavior completely in LMS, allowing for future changes while preserving backward LMS compatibility.


CDrummond:
Not sure how that's _simpler_ - but if it works...

SamInPgh:
> Not sure how that's _simpler_ - but if it works...

Yeah. My explanation wasn't real clear. It basically just better insulates the algorithm for determining volume control from the client app. Thanks!

PR #1052, opened 2025-03-12
Allow SlimBrowse to support duration

smoothquark:
SlimBrowse currently does not seem to support duration values passed by plugins to Material skin. This, together with a trial trivial change to LMS (https://github.com/LMS-Community/slimserver/pull/1348) allows Material to display track duration and total duration. Works with lms-plugin-tidal and Spotty-Plugin.
However, this commit is totally dependent on LMS passing the duration values via SlimBrowse, and would be useless without it.


smoothquark:
Indeed. Thank you.

PR #1058, opened 2025-03-21
Fix album-style play queue when useGrouping=false

darrell-k:
If useGrouping=false, it was splitting on disc number, but using the current track work/grouping as the title. Use album title in that case, and add second line with the disc subtitle/disc number.

I'm not sure how to implement something like this in queue-page:

https://github.com/CDrummond/lms-material/blob/6b62e8823be19ccab669746d6cc20406442f52ec/MaterialSkin/HTML/material/html/js/browse-resp.js#L1799-L1805

### DIFF
--- a/MaterialSkin/HTML/material/html/js/queue-page.js
+++ b/MaterialSkin/HTML/material/html/js/queue-page.js
@@ -74,11 +74,14 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
     if (!queueAlbumStyle || !artistIsRemoteTitle) {
         artistAlbum = addPart(artistAlbum, buildAlbumLine(i, 'queue'));
         let work = buildWorkLine(i, 'queue');
-        if (work && queueAlbumStyle) {
+        if (lmsOptions.useGrouping && work && queueAlbumStyle) {
             artistAlbum = addPart(work, i.work!=i.grouping ? i.grouping : undefined)+'<br/><div class="pq-gsub">'+artistAlbum+'</div>';
             ws = true;
-        } else if (queueAlbumStyle && i.grouping) {
+        } else if (lmsOptions.useGrouping && queueAlbumStyle && i.grouping) {
             artistAlbum +='<br/><div class="pq-gsub">'+i.grouping+'</div>';
+            ws = true;
+        } else if (i.disccount && i.disccount>1 || i.discsubtitle) {
+            artistAlbum +='<br/><div class="pq-gsub">'+ (i.discsubtitle ? i.discsubtitle : i18n("Disc %1", i.disc)) +'</div>';
             ws = true;
         }
         if (queueContext) {


darrell-k:
A couple of related tweaks:
1. Use Grouping regardless of useGrouping when browsing works (disc numbers/names not presently shown anyway).
2. Add Performance to the album-style work heading.

darrell-k:
QUESTION: For single-disc albums, should we display work/grouping subheadings regardless of the useGrouping setting? (Depending on how you read the help text, it could imply the setting is only for multi-disc albums, anyway.)

MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -74,12 +74,15 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
...
+        if (lmsOptions.useGrouping && work && queueAlbumStyle) {
             artistAlbum = addPart(work, i.work!=i.grouping ? i.grouping : undefined)+'<br/><div class="pq-gsub">'+artistAlbum+'</div>';
             ws = true;
-        } else if (queueAlbumStyle && i.grouping) {
+        } else if (lmsOptions.useGrouping && queueAlbumStyle && i.grouping) {
             artistAlbum +='<br/><div class="pq-gsub">'+i.grouping+'</div>';
             ws = true;
+        } else if (i.disccount && i.disccount>1 || i.discsubtitle) {
+            artistAlbum +='<br/><div class="pq-gsub">'+ (i.discsubtitle ? i.discsubtitle : i18n("Disc %1", i.disc)) +'</div>';
+            ws = true;
```
CDrummond, line 85: No. I think we should **only** be using discsubtitle here - either that or they need to come from a comment. I realise trhe use of comment was a band-aid, so I don't think it should be used here. So, if user has explicitly set a discsubtitle we should use, but not otherwise.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1372,7 +1372,7 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
             let compilationAlbumArtist = undefined;
             let extraSubs = [];
             let browseContext = getLocalStorageBool('browseContext', false);
-            let splitIntoGroupings = lmsOptions.useGrouping && (undefined==parent || undefined==parent.multi || MULTI_GROUP_ALBUM==parent.multi || isWork);
+            let splitIntoGroupings = (lmsOptions.useGrouping || isWork) && (undefined==parent || undefined==parent.multi || MULTI_GROUP_ALBUM==parent.multi || isWork);
```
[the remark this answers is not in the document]
CDrummond: And why  leave 'isWork' at then end too? That's redundant.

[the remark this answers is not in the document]
darrell-k: `isWork` here indicates we're browsing an work-album rather than a full album. So OR is correct. I'll simplify the statement.

In the queue there is no `isWork`, because it's just a sequence of tracks which could have been added via any available route. And `work` will always be returned by the LMS status command if the track has a Work tag.

So without some non-trivial change to LMS to let queue-page know the origin of each entry, the best we can do is follow the preference.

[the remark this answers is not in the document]
CDrummond: So, if the queue preference is an AND then it should also be here - as it was before this change.

MaterialSkin/HTML/material/html/js/queue-page.js
darrell-k: That would lead to this:

image

Isn't this better?

image

MaterialSkin/HTML/material/html/js/browse-resp.js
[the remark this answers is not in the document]
darrell-k: No, because unlike in the queue, in the album listing we have the `isWork` flag to tell us what kind of list the user is viewing - work or album. The idea of the change is that we always use grouping for a work list, because it's not the same as an album list.

MaterialSkin/HTML/material/html/js/queue-page.js
CDrummond: **No** I do not like "Disc N" here.

MaterialSkin/HTML/material/html/js/browse-resp.js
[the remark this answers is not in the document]
CDrummond: I want consistency. If its AND in one place, its AND in both - as it already was. Your change adds an inconsistency. In the queue you have the "i.work" ==> its a work. Yes, I know not 100% the same, but close enough.

MaterialSkin/HTML/material/html/js/queue-page.js
darrell-k: In which case we should find a way to not repeat the album title half way down the list, it's redundant.

CDrummond: No. This is how i designed and like it. Sorry, but I do not want to, and will not, change this.

MaterialSkin/HTML/material/html/js/browse-resp.js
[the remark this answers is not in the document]
darrell-k: I don't think it's close at all. `i.work` just means the track has a WORK tag. When browsing, `isWork` is not about the track, it's about which view is the parent of the list.

The requirement for the `useGrouping` preference came about because (mostly Picard/Musicbrainz) was adding WORK/GROUPING tags which the user didn't necessarily know they had until LMS9.

Anyway, I have an idea on how to resolve this which I'll demonstrate with another commit ASAP - (better to show, not tell) - then you can take a view.

MaterialSkin/HTML/EN/plugins/MaterialSkin/settings/basic.html
```diff
@@ -80,7 +80,10 @@
     [% END %]
 
     [% WRAPPER setting title="PLUGIN_MATERIAL_SKIN_USE_GROUPING" desc="PLUGIN_MATERIAL_SKIN_USE_GROUPING_DESC" %]
-    <input type="checkbox" name="pref_useGrouping" id="pref_useGrouping" [% IF prefs.pref_useGrouping %] checked [% END %] value="1"/>
+    <select class="stdedit" name="pref_useGrouping" id="pref_useGrouping">
+        <option [% IF NOT prefs.pref_useGrouping %]selected [% END %]value="0">[% 'PLUGIN_MATERIAL_SKIN_PREFER_DISC_GROUPS' | getstring %]</option>
+        <option [% IF prefs.pref_useGrouping %]selected [% END %]value="1">[% 'PLUGIN_MATERIAL_SKIN_PREFER_WORK_GROUPING_GROUPS' | getstring %]</option>
+    </select>
     [% END %]
```
darrell-k, line 87: Change the preference to a dropdown. Also see strings.txt changes.

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -1088,10 +1088,11 @@ function browseClick(view, item, index, event, ignoreOpenMenu) {
                         if (data.result.work_id && data.result.composer_id) {
                             browseDoClick(view, {id:"album_id:"+data.result.album_id, work_id:data.result.work_id, performance:data.result.performance, composer_id:data.result.composer_id, title:item.title, subtitle:data.result.artist_name, image:item.image, images:item.images, stdItem:STD_ITEM_ALBUM, fromFav:true}, index, event);
                         } else if (data.result.artist_id) {
-                            let itm = {id:"album_id:"+data.result.album_id, artist_id:data.result.artist_id, title:item.title, image:item.image, stdItem:STD_ITEM_ALBUM, fromFav:true};
+                            let itm = {id:"album_id:"+data.result.album_id, title:item.title, image:item.image, stdItem:STD_ITEM_ALBUM, fromFav:true};
```
darrell-k, line 1091: Don't restrict to artist when displaying an album favourite, in case there isn't a common artist.

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -1088,10 +1088,11 @@ function browseClick(view, item, index, event, ignoreOpenMenu) {
...
                             if (data.result.artist_name) {
                                 itm["subtitle"]=data.result.artist_name;
                             }
+                            itm['multi'] = albumGroupingType(data.result.disc_count, data.result.group_count, data.result.contiguous_groups);
```
darrell-k, line 1095: Set the multi flag using the new albumGroupingType function.

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -2628,6 +2629,9 @@ function browseBuildFullCommand(view, item, act) {
                         }
                         if (loop[i].startsWith("artist_id:") && !item.id.startsWith("album_id:")) {
                             command.params.push(SORT_KEY+ARTIST_ALBUM_SORT_PLACEHOLDER);
+                            if (item.stdItem==STD_ITEM_WORK_COMPOSER && LMS_VERSION>=90003) {
+                                command.params.push("work_id:-1");
+                            }
```
darrell-k, line 2634: Restrict to tracks that are part of works when adding all works for a composer or genre. Also allows LMS to know the items are being added from a works browse. This requires an LMS version check as currently the playlistcontrol command can't handle work_id:-1

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -2979,4 +2986,14 @@ function browseFetchExtra(view, fetchArtists) {
...
+function albumGroupingType(discCount, groupCount, contiguousGroups, parentIsWork) {
+    let multiDisc = lmsOptions.groupdiscs && undefined!=discCount && parseInt(discCount)>1;
+    let multiGroup = undefined!=groupCount && parseInt(groupCount)>1 && (undefined==contiguousGroups || 1==parseInt(contiguousGroups));
+    return multiGroup && (lmsOptions.useGrouping || !multiDisc || parentIsWork)
+        ? MULTI_GROUP_ALBUM
+        : multiDisc
+            ? MULTI_DISC_ALBUM
+            : 0;
+}
+
```
darrell-k, line 2998: New centralised function to determine the grouping type to use.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1156,12 +1156,8 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
...
                     subtitleLinks = "<obj class=\"link-item\" onclick=\"show_year(event, "+i.year+",\'"+i.year+"\', \'browse\')\">" + i.year + "</obj>";
                 }
+                let groupType = albumGroupingType(i.disccount, i.group_count, i.contiguous_groups, isWorksAlbums);
 
-                let groupType = lmsOptions.useGrouping && undefined!=i.group_count && parseInt(i.group_count)>1 && (undefined==i.contiguous_groups || 1==parseInt(i.contiguous_groups))
-                                 ? MULTI_GROUP_ALBUM
-                                 : lmsOptions.groupdiscs && undefined!=i.disccount && parseInt(i.disccount)>1
-                                   ? MULTI_DISC_ALBUM
-                                   : 0;
                 let album = {
```
darrell-k, line 1161: Use new albumGroupingType function.

MaterialSkin/HTML/material/html/js/constants.js
```diff
@@ -234,5 +234,6 @@ const STD_ITEM_ONLINE_ARTIST_CATEGORY = 302;
 
 const MULTI_GROUP_ALBUM = 1;
 const MULTI_DISC_ALBUM = 2;
+const ALWAYS_GROUP_HEADING = 2;
```
darrell-k, line 237: Used to force headings in album style play queue.

MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -74,12 +74,22 @@ function buildArtistAlbumLines(i, queueAlbumStyle, queueContext) {
...
+                } else if (i.grouping) {
+                    // track has a grouping tag
+                    artistAlbum +='<br/><div class="pq-gsub">'+i.grouping+'</div>';
+                    ws = true;
+                }
+            } else if (i.discsubtitle) {
+                // track has a discsubtitle tag
+                artistAlbum +='<br/><div class="pq-gsub">'+i.discsubtitle+'</div>';
+                ws = true;
+            }
```
darrell-k, line 92: Use new albumGroupingType function. Reformat code to make it easier to see what's going on (at least for me!)

MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -108,16 +118,19 @@ function parseResp(data, showTrackNum, index, showRatings, queueAlbumStyle, queu
...
                 }
-                pqContiguousMap.set(parseInt(i['playlist index']), parseInt(i.contiguous_groups));
+                if ( !('disccount' in i) ) {
+                    i.disccount = 1; // might not exist for single disc album
+                }
+                pqGroupingMap.set(parseInt(i['playlist index']), [parseInt(i.disccount), parseInt(i.contiguous_groups), parseInt(i.added_from_work)]);
                 i.isClassical = undefined!=i.isClassical && 1==parseInt(i.isClassical);
                 splitMultiples(i, true);
-                let title = queueAlbumStyle && i.contiguous_groups==1 ? i.title : trackTitle(i);
+                let title = queueAlbumStyle && albumGroupingType(i.disccount, ALWAYS_GROUP_HEADING, i.contiguous_groups, i.added_from_work)==MULTI_GROUP_ALBUM ? i.title : trackTitle(i);
```
darrell-k, line 133: Use the new function to assist in setting title.

MaterialSkin/HTML/material/html/js/queue-page.js
```diff
@@ -508,7 +521,10 @@ var lmsQueue = Vue.component("lms-queue", {
                 var index = playerStatus.current["playlist index"];
                 if (undefined!=index && index>=0 && index<this.items.length) {
                     var i = playerStatus.current;
-                    var title = this.$store.state.queueAlbumStyle && pqContiguousMap.get(parseInt(index))==1 ? i.title : trackTitle(i);
+                    var discCount = pqGroupingMap.get(parseInt(index))[0];
+                    var contiguousGroups = pqGroupingMap.get(parseInt(index))[1];
+                    var addedFromWork = pqGroupingMap.get(parseInt(index))[2];
+                    var title = this.$store.state.queueAlbumStyle && albumGroupingType(discCount, ALWAYS_GROUP_HEADING, contiguousGroups, addedFromWork)==MULTI_GROUP_ALBUM ? i.title : trackTitle(i);
```
darrell-k, line 527: Use the enhanced map and new function

MaterialSkin/Plugin.pm
```diff
@@ -885,8 +885,18 @@ sub _cliCommand {
             }
             if ($albumId) {
                 $request->addResult('album_id', $albumId);
-                if ($artistName) {
-                    $request->addResult('artist_name', uri_unescape($artistName));
+                if ( $artistName = uri_unescape($artistName) ) {
+                    utf8::decode($artistName);
+                    $request->addResult('artist_name', $artistName);
+                }
```
darrell-k, line 891: Sort out character encoding (also see same for performance, below).

MaterialSkin/Plugin.pm
```diff
@@ -885,8 +885,18 @@ sub _cliCommand {
...
+                my $albumsRequest = Slim::Control::Request->new( undef, [ 'albums', 0, 1, "album_id:$albumId", "tags:2q" ] );
+                $albumsRequest->execute();
+                if ($albumsRequest->isStatusError()) {
+                    $log->error($albumsRequest->getStatusText());
+                } else {
+                    $request->addResult('disc_count', @{$albumsRequest->getResult('albums_loop')}[0]->{'disccount'});
+                    $request->addResult('group_count', @{$albumsRequest->getResult('albums_loop')}[0]->{'group_count'});
+                    $request->addResult('contiguous_groups', @{$albumsRequest->getResult('albums_loop')}[0]->{'contiguous_groups'});
```
darrell-k, line 899: Use albumsQuery to get derived data. This fixes grouping when browsing favourited albums.

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -2979,4 +2986,14 @@ function browseFetchExtra(view, fetchArtists) {
     });
 }
 
+function albumGroupingType(discCount, groupCount, contiguousGroups, parentIsWork) {
```
CDrummond, line 2989: Can you move this function into `uitils-deferred.js` - as its not browse specific (queue also uses this).

PR #1066, opened 2025-04-19
Allow user control of max albums shown in Now Playing (MAI artist tab)

darrell-k:
For users with large collections: see https://forums.lyrion.org/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1764547#post1764547

### DIFF
--- a/MaterialSkin/HTML/material/html/js/nowplaying-functions.js
+++ b/MaterialSkin/HTML/material/html/js/nowplaying-functions.js
@@ -4,7 +4,7 @@
  * Copyright (c) 2018-2024 Craig Drummond <craig.p.drummond@gmail.com>
  * MIT license.
  */
-const NP_MAX_ALBUMS = 50;
+const NP_MAX_ALBUMS = Math.max(50, lmsOptions.pagedBatchSize);
 const NP_MAX_TRACKS = 50;
 
 function  nowPlayingHeader(s) {


CDrummond:
No, sorry - you are using 1config item to control something else. Why does this even need to be configurable?

darrell-k:
Well, the help text for that says "Number of initial items to fetch (and display) when listing contents of some lists (such as playlist contents). Scrolling list will then cause another batch of items to be fetched, etc." and I thought the "etc" covered it. Maybe not.

As I said in the forum thread, I was concerned about the size of the cache.

What would you propose?

CDrummond:
I propose nothing - the code is fine as is. It was only **ever** meant to get the first few releases - which is what it does. From my perspective, nothing needs to change.

darrell-k:
The issue was raised by a user who carefully sequences his release types, so in the main browse the first 50 albums shown are not the 50 shown in the MAI view, because of course the main listing grabs much more and then resequences internally. 

If you don't like sharing the preference, would you consider a separate pref for this?

CDrummond:
No! So **one** user has an issue? This list is ordered by year then filtered. There is **nothing** that need changing. Its only a quick list of the first 50 releases of an artists - its **deliberately** capped at a small number, no view recycler is used, etc. Sorry, but no this does not need changing.

darrell-k:
Every issue starts with one user...

CDrummond:
And I will only make changes that I want and agree with. As stated there is a **reason** why the limit of 50 is chosen.

darrell-k:
I'm only trying to discuss it. Your point about the recycler is noted. Maybe I'll look at using the Material setting for release type sequence in LMS if it's available, so that the first 50 albums will be the same as in the main browse list.

CDrummond:
Please don't. It's the first 50 releases regardless of type - sorted by year. The actual artist album list can be sorted differently.

darrell-k:
Maybe consider removing the category headings from the MAI album list, then. That was where the initial user confusion came from: seeing the category with obviously missing releases.

PR #1068, opened 2025-04-19
Create release categories for user-defined roles

darrell-k:
See https://forums.lyrion.org/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1764606#post1764606

### DIFF
--- a/MaterialSkin/HTML/material/html/js/browse-resp.js
+++ b/MaterialSkin/HTML/material/html/js/browse-resp.js
@@ -1139,6 +1139,9 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                             group = "APPEARANCE_BAND";
                         } else if (roles.has(COMPOSER_ARTIST_ROLE)) {
                             group = "COMPOSITION";
+                        } else if (!(roles.size===0)) {
+                            let firstRole=[...roles][0];
+                            group = lmsOptions.userDefinedRoles[firstRole].name;
                         }
                     }
                     showRoles = new Set([...showRoles, ...roles]);
@@ -1264,6 +1267,8 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                         resp.subtitle=i18np("1 Appearance", "%1 Appearances", resp.items.length)+appearanceSuffix(releaseType);
                     } else if (releaseType=="COMPOSITION") {
                         resp.subtitle=i18np("1 Composer Album", "%1 Composer Albums", resp.items.length);
+                    } else if (releaseType!="ALBUM") {
+                        resp.subtitle=releaseType;
                     } else if (lmsOptions.supportReleaseTypes) {
                         resp.subtitle=i18np("1 Release", "%1 Releases", resp.items.length);
                     } else {


CDrummond:
But a "user role" is **not** a "release type".

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1139,6 +1139,9 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                             group = "APPEARANCE_BAND";
                         } else if (roles.has(COMPOSER_ARTIST_ROLE)) {
                             group = "COMPOSITION";
+                        } else if (!(roles.size===0)) {
```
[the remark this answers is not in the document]
darrell-k: Not entirely sure. Or even if it is really needed, but I thought belt and braces.

darrell-k:
> But a "user role" is **not** a "release type".

Neither are track artist, conductor, composer, band, but the're already there.

MaterialSkin/HTML/material/html/js/browse-resp.js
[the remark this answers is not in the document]
CDrummond: But it does not read easily. Sorry, but `>0' is what you mean, so just use that.

CDrummond:
Yeah, I'm looking at that (sorry did not full expand to see what the change was doing) - and that does not make sense to me. Perhaps they should all be under "Appearance(s)"??? As in I can't see the need for "Lyricist" being a grouping along with "EP", "Compilation", etc. They just do not belong together.

darrell-k:
If someone is specifically interested in Lyricist, such that they've created the tag and a user-defined role for it, it's every bit as valid for them as conductor etc, and like those, if you don't tag it you'd never see it.

darrell-k:
And categorising composer or lyricist (or some other roles that people now use, engineer, for example) as an "appearance" is just not right.

CDrummond:
But they are also **not** release types! Hence the mixing is incorrect. Yes, I agree if they've been tagged as a Lyricist then you'd want to find them. But I do not agree that "Lyricist" should be mixed with "Albums", "EPs", "Singles" - that's a different grouping _type_.

Why is "Lyricist" not an _appearance_?

darrell-k:
It's an example of a role who not on the stage or in the studio. 

If you only want actual release types then Appearances (which is TRACKARTIST)  shouldn't be there either. 

Applying release types only to ARTIST/ALBUMARTIST roles and then having extra categories for other roles is how it's worked for a long time.

CDrummond:
I agree that "Appearances" also does not make sense - so you are correct there.

Perhaps "Appearances" should just be in the main "..." menu, and that list of "Appearances" should have "Track artist", "Composer", "Lyricist", etc. groupings.

MaterialSkin/HTML/material/html/js/browse-resp.js
```diff
@@ -1264,6 +1267,8 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG
                         resp.subtitle=i18np("1 Appearance", "%1 Appearances", resp.items.length)+appearanceSuffix(releaseType);
                     } else if (releaseType=="COMPOSITION") {
                         resp.subtitle=i18np("1 Composer Album", "%1 Composer Albums", resp.items.length);
+                    } else if (releaseType!="ALBUM") {
+                        resp.subtitle=releaseType;
```
CDrummond, line 1271: No. This is different to all other sub-headings here.

darrell-k:
That would be quite a change (from a user point of view). Material has so far tracked Default Skin in this area. There would be complaints, I'm sure.

Presumably you'd also have to include the albums in question in the *actual*  release type categories. 

Personally I think it would be a real step backwards, for no good reason. It's really nice to see an artist's differing roles in a library laid out like presently, without having to drill down into other menus.

But if you go ahead with this, please don't call the "..." menu option Appearances, it's misleading. "Artist contributions by role" perhaps?

MaterialSkin/HTML/material/html/js/browse-resp.js
darrell-k: Because the user provides their own preferred text for user-defined roles in the LMS settings, so no need for i18n.

CDrummond:
A, Material is **not** Default so changes are to be expected, where they make sense - not saying that's the case here, but stating the fact that's is different as being an issue is invalid.

Well they make an "appearance" as a "Lyricist". But, I'm not sure what the best heading would be.

MaterialSkin/HTML/material/html/js/browse-resp.js
CDrummond: No. **All** others are "1 XXX" / "N XXX" - that is the difference. So, I'm thinking these should all be "Appearance (ROLE)"  - and then "1 Appearance" / "N Appearances". Which is then consistent with the other "Appearance" categories.

CDrummond:
Also, if you use "roles" as a category here then release-types are mixed in. e.g. You could have an artist with Albums, EPs, and singles whose also the Lyricist on other EPs and Singles - but these last two are grouped under Lyricist.

Material already provides access to the (e.g.) "Lyricist" role in the "..." menu. Selecting that will allow grouping via actual release-type.

This change also only groups under 1 role - but if an artist has multiple-roles on a release only 1 would be shown.

So, again, I fail to see what the issue is with the current Material code. Currently Material groups by release_type, Appearance, Composition, Conductor, Band/Orchestra, and Composer - and then allows filtering by Role. Now perhaps these Appearance, Composition, Conductor, Band/Orchestra, and Composer headings should be moved to the "..." menu - to me that would be the consistent change.

darrell-k:
> A, Material is **not** Default so changes are to be expected, where they make sense - not saying that's the case here, but stating the fact that's is different as being an issue is invalid.

I made exactly that point in the forum thread when discussing differences between Material and Default. I was just pointing out that up to now the two have tracked quite closely in this area (there are already differences, of course, for example when an album has the compilation flag set, Default Skin ignores the release type and puts in the Compilations category), so there would be confusion if it suddenly diverged.

CDrummond:
Please try the `1068-roles` branch. This _should_ move all 'role' filters to the '...' menu - as that's what these are, filters not groups. This then allows role:drummer to group on EP, Single, etc.

CDrummond:
Aah! Just noticed that that branch breaks getting user-defined roles - as in only those of the album list will show. But means you can go from Guitarist -> Vocals, whereas before you could. Still, it shows the idea that the role-filters should be in "..." and release_type used for grouping. Will work on branch (and restore the removed code).

darrell-k:
> Also, if you use "roles" as a category here then release-types are mixed in. e.g. You could have an artist with Albums, EPs, and singles whose also the Lyricist on other EPs and Singles - but these last two are grouped under Lyricist.
> 
> Material already provides access to the (e.g.) "Lyricist" role in the "..." menu. Selecting that will allow grouping via actual release-type.
> 
> This change also only groups under 1 role - but if an artist has multiple-roles on a release only 1 would be shown.
> 
> So, again, I fail to see what the issue is with the current Material code. Currently Material groups by release_type, Appearance, Composition, Conductor, Band/Orchestra, and Composer - and then allows filtering by Role. Now perhaps these Appearance, Composition, Conductor, Band/Orchestra, and Composer headings should be moved to the "..." menu - to me that would be the consistent change.

At present, if the artist role is not one of the 6 pre-defined LMS roles, the release will end up in the "Albums" category regardless of its release type.

It is only ARTIST and ALBUMARTIST roles that are considered for inclusion in an actual Release Type category - hence  `ARTIST_ROLES` . It was this quirk which was noticed by the user: a release which they tagged as releasetype=SINGLE ended up in the Albums category when browsing the lyricist.

So in the existing code, for this example, clicking on "Lyricist" in the "..." menu would put the album on which the contributor is lyricist (which has a release type of SINGLE) in the "Albums" category again.

Again, in the existing code, where the artist has multiple roles on a release, if one of those roles is in ARTIST_ROLES, the release will go into an actual release type category, otherwise it will go into the first category matched by the `if... else if...` structure in the code (with the "Album" category as default).

The exception to this is the change I made in a previous PR, whereby the "from menu" role is prioritised if the artist is being browsed from a role-specific menu.

So, in summary, all this change is doing is treating user-defined roles the same as the pre-defined (non-ARTIST_ROLES) roles.

Having said that, I'm starting to see the attraction of your suggestion that all releases featuring the artist should initially be shown cleanly by their actual release type, with the "..." options to browse for specific (non-ARTIST_ROLES)  roles. Yes, as I said before, this might cause some user confusion, but on reflection it might be worth it. 

Leave it with me and I'll amend the PR along those lines.

darrell-k:
> Aah! Just noticed that that branch breaks getting user-defined roles - as in only those of the album list will show. But means you can go from Guitarist -> Vocals, whereas before you could. Still, it shows the idea that the role-filters should be in "..." and release_type used for grouping. Will work on branch (and restore the removed code).

Agh! messages crossed. I'll wait for your further changes on the branch, then have a look.

CDrummond:
Try branch now. Only issue is with 'Appearances' - i.e. where an artist appears on an album but is not main artist. Currently if 'roles' returns '6' for an artist then I assume 'Appearances' can be added?

darrell-k:
I made the following changes:

```
--- a/MaterialSkin/HTML/material/html/js/browse-resp.js
+++ b/MaterialSkin/HTML/material/html/js/browse-resp.js
@@ -1099,14 +1099,12 @@ function parseBrowseResp(data, parent, options, cacheKey, parentCommand, parentG

```
To ignore roles when setting the group, otherwise a release is lumped into ALBUMS regardless of its release type if the browsed artist role on the release is not in ARTIST_ROLES, which was the originally reported problem:
```
                 }
                 let group = "ALBUM";
                 let roles = new Set(isEmpty(i.role_ids) ? [] : splitIntArray(i.role_ids));
-                if (intersect(ARTIST_ROLES, roles).size>0 || roles.size==0) {
                     if (lmsOptions.groupByReleaseType>0) {
                         let isCompilation = undefined!=i.compilation && 1==parseInt(i.compilation) && (undefined==i.release_type || i.release_type.toUpperCase()=="ALBUM");
                         group = isCompilation ? "COMPILATION" : undefined==i.release_type ? "ALBUM" : i.release_type.toUpperCase();
                     }
-                }

``` 

resp.ignoreRoles is a set... (still trying to get my head around what this is for)
```
                if (0==roles.size || !roles.has(TRACK_ARTIST_ROLE)) {
-                    resp.ignoreRoles(TRACK_ARTIST_ROLE);
+                    resp.ignoreRoles.add(TRACK_ARTIST_ROLE);
                 }

```

Still testing, but on the "..." entries, would it be good to put separators above and below the list of role filters, and also include ARTIST/ALBUMARTIST/TRACKARTIST (together or separately, I don't know yet)?

CDrummond:
> resp.ignoreRoles is a set... (still trying to get my head around what this is for)

When browsing My Music / Artist / X / Vocalist - I don't want 'Vocalist' in the '...' menu, as you're already browsing by that role.

...changes applied.

darrell-k:
I'm not seeing entries in the "---" menu for the server-defined roles.

CDrummond:
Not sure what you are expecting, but `browseGetRoles` in `browse-functions.js` calls `roles` to get all roles of an artist. These are then added to the menu. So, look at that to see what its doing, and if you think something is missing.

CDrummond:
I've applied your changes. I changed the `multipleRoles` to need 2 or more roles - otherwise its not multiple!

Can this pull request now be closed?

PR #1122, opened 2025-11-03
Drill down favorites recursively.

michaelherger:
Haha... just saw that you've started working on this, too... But anyway: as I've already done it, I submit it. As you'll see I get the list of favorites directly, not using the CLI. This should be save enough.


CDrummond:
Thanks!

PR #1138, opened 2025-11-14
Add missing German translations

chrober:
@CDrummond, I wasn't sure about `Remote Playlist` (therefore the draft status of this PR).

Does "remote" refer to playlists originating from Remote Libraries (which currently translates to `Netzwerk-Musiksammlungen`)?

Note: `Netzwerk-Musiksammlungen` refers to music libraries available over a network. While "Netzwerk" is not a literal translation of "remote", this is the term currently used in the German UI for Remote Libraries.


chrober:
> These are, I assume, from online services. I say assume as the remote flag is from LMS.

Maybe "Online Musiksammlungen" and "Online Wiedergabeliste" would fit better... let me think twice about it.

PR #1144, opened 2025-11-22
Add translatable string token for "Presets"

michaelherger:


CDrummond:
I thought 'Presets' was added as an example only? And should actually be removed?

Not sure I see the point in providing it as a real list.

michaelherger:
Sure, feel free to remove it. Most users likely have the same in their favorites anyway.

CDrummond:
Closing then, will remove example code - as its no longer needed.

PR #1170, opened 2026-01-08
Allow plugins/3rd parties to trigger a home extra update

michaelherger:
Don't only update the extra configuration, but the actual items, too. Some plugins change their content, like the random play does. But we want them to be able to update without the user having to interact with the menu. Eg. 1001 Albums would update once a day. Currently the user has to either navigate away from home and back, or re-load the page in order to pick up the latest album. With this change the plugin will be able to push an event to the frontend.

### DIFF
--- a/MaterialSkin/HTML/material/html/js/server.js
+++ b/MaterialSkin/HTML/material/html/js/server.js
@@ -823,6 +823,7 @@ var lmsServer = Vue.component('lms-server', {
             this.cancelHomeRefreshTimer();
             this.homeRefreshTimer = setTimeout(function () {
                 lmsCommand("", ["material-skin", "home-extra-3rdparty"]).then(({data}) => {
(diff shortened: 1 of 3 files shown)


MaterialSkin/HTML/material/html/js/server.js
```diff
@@ -823,6 +823,7 @@ var lmsServer = Vue.component('lms-server', {
             this.cancelHomeRefreshTimer();
             this.homeRefreshTimer = setTimeout(function () {
                 lmsCommand("", ["material-skin", "home-extra-3rdparty"]).then(({data}) => {
+                    bus.$emit('refresh-home');
```
CDrummond: I think this should be emitted in store.js after the home3rdPartyExtraLists is updated - but I'll make that change.

PR #1198, opened 2026-02-22
Processing-User-ID

Sveninndh:
With HomeExtras, you can now override the number of elements that are scrolled; the default value is 10.

The maximum number of Lyrion users is now 9, which should always be sufficient in practice.

Furthermore, the ‘user_id’ is now also passed on to ‘homeExtras’, because at the time this home page is built, a plugin has not yet had any user action and therefore does not yet know which user is requesting data.


CDrummond:
Sorry, but no. This is not how I would want this implemented. What would be required:

1. Another "LMS Users" plugin that will allow the creation of users. This can have the username, avatar, etc. The LMS user would then assign Qobuz, etc, accounts to this LMS-user. Or the Qobuz plugin can do this, whichever makes more sense
2. Material would check for the existence of this plugin and show a user-selector somewhere (probably in the hamburger menu). When a "UserID" is selected, this will be used in all calls.

Having a hard-coded list of 9 user-ids in Material is a no-go. This is all code specific to your Qobuz plugin.

> With HomeExtras, you can now override the number of elements that are scrolled; the default value is 10.

This is not related to user-IDs, so should not be in this merge request. Material will set the number of scroll elements, not the plugin - for consistency, and I don't see why this should change.

--------------

As a general note please discuss ideas before creating such requests.

CDrummond:
@Sveninndh I take it from the lack of replies that you are not interested in my proposed solution?

Sveninndh:
No, you can't really say that. I did mention in the forum that I’m currently very busy with other things. What you’re suggesting and want to demonstrate in the dev version isn’t something you can just test in five minutes. Besides, I never claimed that your solution is better or worse than mine. Maybe my solution requires a little less effort—at least for now and as things stand for me today—since it works for my needs. And when I look at the feedback I’m getting on this, I have to admit that nobody really cares. I wanted to briefly release my modified version of MaterialSkin, but the community’s only insurmountable problem seemed to be the 60.x.x serial number—it would confuse users.
Sorry, what is this nonsense? I assume the number of users who treat themselves to a subscription with multiple accounts is negligible. So there’s clearly zero interest there—so why should we continue wasting our precious time on it? I’ve solved my problem and I can live with performing an update to MaterialSkin every now and then; it doesn’t have to be with every version update.
And I’ll also stop submitting pull requests for Lyrion in the future, because the effort is getting out of hand—the core developers keep coming up with new demands and requests and then don’t implement any of them (though maybe they will eventually). From their perspective, it’s perhaps understandable; from mine, it’s just inefficient. I update Lyrion maybe four times a year. The few changes I need to make so that MaterialSkin runs really smoothly take me maybe 15 minutes a year. Getting that pull request approved has already cost me countless hours without leading to any results.
Sorry for venting to you, but maybe this helps you understand how I feel about it.

PR #1199, opened 2026-03-09
Suppress duplicate composer in work-grouped track listings

mspertus:
## Summary
- When tracks are grouped by classical work, the composer name already appears in the group header (e.g. "Igor Stravinsky • Ebony Concerto"). This removes the redundant composer from individual track subtitles within those groups.
- Applies to both the **browse** (album track list) and **queue** views.
- Non-classical albums and tracks outside work groups are unaffected.

## Details
- **queue-page.js**: Extracts `isMultiGroup` to avoid a duplicate `albumGroupingType()` call, then passes `useComposerTag: false` to `buildArtistLine()` when the track is in a composer+work group.
- **browse-resp.js**: Builds a composer-free artist line variant alongside the normal one during the track loop. After the loop, when group headers are confirmed to be shown (`willShowGroupHeaders`), uses the no-composer variant for tracks within groups.

## Forum report
https://forums.lyrion.org/forum/user-forums/logitech-media-server/1737175-...?p=1813835#post1813835

## Test plan
- [ ] Browse a work-grouped album (e.g. "Ebony Concerto") — group headers show composer, individual tracks do not
- [ ] Add work-grouped tracks to the queue — same behavior
- [ ] Browse a non-classical album — no change in behavior
- [ ] Browse an album with mixed work/non-work tracks — only grouped tracks suppress composer

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MaterialSkin/HTML/material/html/js/queue-page.js
+++ b/MaterialSkin/HTML/material/html/js/queue-page.js
@@ -182,12 +182,18 @@ function parseResp(data, showTrackNum, index, showRatings, queueAlbumStyle, queu
                         grpHeaderNames.add(artist);
                     }
                     if (i.band && lmsOptions.showBand && useBand(i)) {
-                        if (i.bands) i.bands.forEach(function(b) { grpHeaderNames.add(b); });
-                        else grpHeaderNames.add(i.band);
+                        if (i.bands) {
+                            i.bands.forEach(function(b) { grpHeaderNames.add(b); });
+                        } else {
+                            grpHeaderNames.add(i.band);
+                        }
                     }
                     if (i.conductor && (i.work || (lmsOptions.showConductor && useConductor(i)))) {
-                        if (i.conductors) i.conductors.forEach(function(c) { grpHeaderNames.add(c); });
-                        else grpHeaderNames.add(i.conductor);
+                        if (i.conductors) {
+                            i.conductors.forEach(function(c) { grpHeaderNames.add(c); });
+                        } else {
+                            grpHeaderNames.add(i.conductor);
+                        }
                     }
                 } else if (isAlbumHeader) {
                     grpHeaderNames = undefined;


darrell-k:
I'll give it a quick test.

CDrummond:
Is this ready for merge?

darrell-k:
A work can only have one composer, a limitation of the LMS data model. Not sure what happens if a track or group of tracks have a work tag and multiple composer tags. The scanner might create multiple works, one for each composer, or just ignore all but the first composer for the purposes of work creation. I can't remember. 

Anyway, I'll try to remember to test the latest changes tomorrow.

mspertus:
Alright, it looks good in my testing. Appreciate if you could test again @darrell-k . 
Thanks, Mike

MaterialSkin/HTML/material/html/css/style.css
```diff
@@ -3123,6 +3123,14 @@ div.v-subheader {
  font-size:var(--std-font-size) !important;
 }
 
+.browse-header .v-list__tile__title .link-item {
+ color:var(--text-color) !important;
```
CDrummond: Sorry, but no. I don't like this change. All headers use the active colour to distinguish them from tracks, etc. This is make these items different to the rest.

darrell-k:
@mspertus , @CDrummond I would suggest an extra improvement which would be to reverse the sequence of PERFORMANCE and GROUPING in the track listing and queue subheadings, so that GROUPING precedes PERFORMANCE when both are present.

This is because in that case GROUPING is a subset of the WORK, so those are better kept together rather than be separated by PERFORMANCE.

darrell-k:
@mspertus looks OK on the surface, please push the changes so that I can test.

darrell-k:
> No, you're correct - too much change. **Just use all the artists from the first track**, that will be good enough until I've designed and implemented the change to LMS that is necessary.

I'm still seeing the first artist in the header and the second against each track in the Grieg Songs example above. All tracks are tagged with both artists.

darrell-k:
This looks good to me now.

mspertus:
@CDrummond Just want to make sure you saw the above comment. I wouldn't blame you for tuning out the blow-by-blow of the discussion 

Mike

PR #1217, opened 2026-05-03
Match queue album-style track tile height to album-header slice

mspertus:
## Problem

In the queue's album-style view, every album header tile contains the album/work text plus a 34 px slice at the bottom that holds the first track of the album. The slice provides 34 px of content (the wrapper border belongs to the tile around it). A standalone subsequent-track tile, by contrast, is sized at 34 px **total** — 33 px content + 1 px wrapper `border-bottom`.

Net effect: the first-track row in every album header renders 1 px taller than every subsequent-track row. The asymmetry is structural to the slice trick and applies to all album-header tiers.

In the picture below, I've pasted the second track in a group (Begin the Beguine) next to the first track,to show that its height is smaller.
image

## Fix

Bump `LMS_ALBUM_QUEUE_TRACK` and `--pq-album-track` from 34 → 35. A standalone track tile now renders 34 px content + 1 px wrapper border = 35 px visible row, matching the slice's 34 px content + 1 px shared wrapper border = 35 px effective row. I bumped the track size rather than decreasing the header size because that had a smaller blast radius and appeared to allow for better centering within a track (tested by examining whether the vdots in the track have the same number of pixels above and below)

The constants stay in lockstep so the RecycleScroller's per-item `size` field continues to match the rendered DOM height.

Here is an updated screenshot after the fix

image

## Scope

Both values are referenced only inside `.lms-list-album` — the album-style queue view. Five usages total:
- `constants.js` definition
- `parseResp` size assignment for non-header items
- `--pq-album-track` definition
- `.lms-list-album .pq-track .v-list__tile` height
- `.lms-list-album .pq-current-indicator` bottom positioning (play indicator shifts by 0.5 px — imperceptible)

Not affected: flat (non-grouped) queue list, browse view, search results, now-playing, settings.

## Verified

Tested in album-style queue with multiple album types in Windows (Chrome), iPad (LyrPlay, Safari, Chrome, and Firefox), and Galaxy S22. The first-track row now renders the same visible height as subsequent rows.  The screenshot and exact measurements were made on Windows. Note that there are some small preexisting differences due to how pixels get mapped to each display, but these are improved or not made worse by this change.

The recycle scroller, highlight, and jump-to-current are all tested as well. Jump-to-current scroll position still lands correctly because the album-style branch sums per-item `.size` directly.


CDrummond:
Thanks!

PR #1218, opened 2026-05-03
Show release name on its own subheading line in queue work header

mspertus:
## Problem

While the queue work header currently puts both the contributors and the release name on the same line,  I was thinking that if they were on separate lines, it might fit better, be more consistent with the release browser, and I found having both artists and release titles in the same list felt a little unnatural to me. This screenshot shows the current version

image

## Fix

Split the work-header subheading into two lines: artist on one, release on the next. The work title stays on the first line of the header (e.g., "Composer · Work") as shown in this screenshot
image

## Layout changes

- New `LMS_WORK_QUEUE_HEADER` = 96 / `--pq-work-header:96px` tile tier for headers that have a third subheading line. Plain album header (68 px) and group header (82 px) tiers are unchanged.
- New `.pq-work-header` CSS class with title `margin-top:27`, time `margin-bottom:-62`, menu `margin-bottom:-38`, `.pq-album-header` `padding-top:8` — calibrated so the first-track slice geometry matches the existing 82 px group-header tier and the 3-dot menu / duration align with the title baseline.
- New `.pq-current-first-track:is(.pq-work-header)` gradient at 63.5% so the bright current-track color transitions at the partial-width hline.
- `LMS_ALBUM_QUEUE_TRACK` / `--pq-album-track` 34 → 35 so a standalone track tile (34 content + 1 wrapper border) renders the same visible row height as a first-track inside any album-header tier (slice's 34 content + 1 wrapper border = 35).

## JS plumbing

- `buildArtistAlbumLines` returns a third tuple element `isWork`, set only for the 3-line work case.
- `parseResp` threads `isWork` through to the rendered item as `isWorkHeader`, mutually exclusive with `isGrpHeader`.
- Template binds `'pq-work-header': item.isWorkHeader` on the v-list-tile alongside the existing `pq-grp-header` binding.
- Size ternary picks `LMS_WORK_QUEUE_HEADER` when `isWorkHeader` is set, preserving the JS-CSS height invariant the RecycleScroller relies on.

## Testing

Tested on various Windows, iOS, Android configurations. There are some small pre-existing differences by platform likely due to pixel mapping differences, but highlighting, virtual scroll, and Jump-to-work use the new constants and no regressions noted

## Relation to [Match queue album-style track tile height to album-header slice](https://github.com/CDrummond/lms-material/pull/1217)
This PR includes the same change as PR 1217 to facilitate proper layout. @CDrummond , recommend you merge that one first, and there will be no conflict, so the merge should go smoothly. However, if you prefer, I can remove that change from this PR for a cleaner git history.


CDrummond:
I've fixed the merge conflict - due to merging your previous pull first. To be honest, not sure why git had an issue with that.

Can you confirm the current track indicator displays correctly for the first track in a work? It's not shown on any screenshot.

MaterialSkin/HTML/material/html/js/constants.js
```diff
@@ -65,6 +65,7 @@ const LMS_LIST_3LINE_ELEMENT_SIZE = 70;
 const LMS_ALBUM_QUEUE_HEADER = 68;
 const LMS_GROUP_QUEUE_HEADER = 82;
 const LMS_WORK_QUEUE_HEADER = 96;
+const LMS_GROUP_QUEUE_HEADER = 82; // Group header has 2 lines before track - album name/group name
```
mspertus, line 68: This looks like a duplicate of line 66. Can you please delete one of them before merging?

mspertus:
Hi @CDrummond ,
Here is a screenshot of the queue when the current track is the first track in a work. Please let me know if that is what you need.
image

Note my comment above about [this commit](https://github.com/CDrummond/lms-material/pull/1218/commits/2778eac426e9a02106f50ffa6a7b533ae97a2b85) having introduced a duplicate line. Can you please delete before merging?

Thanks, Mike

MaterialSkin/HTML/material/html/js/constants.js
mspertus: @CDrummond

CDrummond:
That's perfect, thanks.

Don't worry about the duplication - that's my bad. Will fix after merge.

PR #1229, opened 2026-05-25
Show work header for single-work albums.

mspertus:
This addresses LMS-Community/slimserver#1572 by ensuring a work header appears for each work regardless of whether it is the only work in a release. This is both helpful because useful info like composer that would otherwise be missing is now shown and because showing the work header for any work is more consistent.

Before the change, this single-work album was missing the work header
image

After the change, the work header is there, showing that Bach is the composer for the work
image

which agrees with what it is in multi-work albums
image

Since adding a "single work album" flag seemed too much like an *ad hoc* hack when the real issue is whether it has works at all, the implementation (a joint effort of me and Claude Code) now says that if there is any groupings they get a group header while remaining careful to avoid inadvertently creating things like a disk header if there is only one disk. This does result in two behavior changes
* If there is an explicit grouping tag, the group is always displayed
* If there is an album that has some tracks that are works but some that are not, the work headers are now consistently displayed with the non-work tracks in an "Others" header

My sense is that those are likewise good refinements that enhance consistency, but others may well disagree, so calling out explicitly.

### DIFF
--- a/MaterialSkin/HTML/material/html/js/browse-resp.js
+++ b/MaterialSkin/HTML/material/html/js/browse-resp.js
@@ -1484,7 +1484,13 @@ function parseBrowseResp(data, parent, options, cacheKey) {
             let compilationAlbumArtist = undefined;
             let extraSubs = [];
             let browseContext = getLocalStorageBool('browseContext', false);
-            let splitIntoGroupings = undefined==parent || undefined==parent.multi || MULTI_GROUP_ALBUM==parent.multi;
+            // Attempt grouping whenever the album isn't already in multi-disc mode. The per-track
+            // logic produces no groupings for albums without work/grouping tags, and the gates
+            // further down (groupings.size>0) suppress header output in that case. Excluding
+            // MULTI_DISC_ALBUM preserves disc-based headers and the disc-choice prompt for
+            // multi-disc single-work albums (e.g. a 3-CD St Matthew Passion) where disc grouping
+            // is the more useful structure.
+            let splitIntoGroupings = undefined==parent || MULTI_DISC_ALBUM!=parent.multi;
 
             for (let idx=0, loop=data.result.titles_loop, loopLen=loop.length; idx<loopLen; ++idx) {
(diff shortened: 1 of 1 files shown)


mspertus:
@CDrummond @krefson @darrell-k

MaterialSkin/HTML/material/html/js/browse-functions.js
```diff
@@ -1931,6 +1931,17 @@ function browseItemAction(view, act, origItem, index, event, slimBrowseBaseActio
                         discs.push(loop[i]);
                     }
                 }
+                // Only one sub-unit (e.g. single-work album) — no meaningful choice, just play the album.
+                if (discs.length<=2) {
```
CDrummond: Not an issue with this pull request, but... Does this mean that the multi-disc / multi-group icon is being shown in browse? If so then that should also be fixed - as the idea of that icon is to inform the user that they _will_ be prompted for which disc/group to play.

mspertus: Looks like I introduced a problem with this change (I missed because it was suppressed in my testing due to a stale browser cache. I will do all my testing in incognito windows going forward). Thanks, @CDrummond! Let me fix and update the PR. I will also test the preexisting behavior to see if it is what you want.

mspertus:
I've updated the logic to reflect that the presence of an icon on the album indicates that the user will be prompted for a choice (per @CDrummond's [comment](https://github.com/CDrummond/lms-material/pull/1229#discussion_r3301825800)). In the following screenshot:
* AC/DC - Regular album plays with no prompt, no icon
* Bach for Harpsichord - Many works on one disk, prompts for work, notes icon
* Brandenburg Concertos - Many works, multiple disks, prompts for work, notes icon
* St. Matthew's Passion - One work, many disks, prompts for disk, disks icon
* Goldberg Variations - One work on a single disk, plays with no prompt, no icon
* BBC Sessions - Multi-Disk, no works, prompts for disk, disks icon

image

CDrummond:
Thanks!

PR #1234, opened 2026-06-14
Forgotten replacement, no impact, just optimized and clearer code

Sveninndh:
In lines 2113 and 2118, $index is already defined and contains the request parameter '_P2'. $index should therefore be used in the subsequent code as well; hence the change in line 2261.

### DIFF
--- a/MaterialSkin/Plugin.pm
+++ b/MaterialSkin/Plugin.pm
@@ -2258,7 +2258,7 @@ sub _handleHomeExtraCmd {
                 my $id = $extra->{id};
 
                 my $args = { 
-                    index    => $request->getParam('_p2') || 0,
+                    index    => $index,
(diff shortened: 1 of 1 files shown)


Sveninndh:
Craig, I’m still struggling with this Git—or rather, GitHub—system. I find it annoying that if you make minor corrections after submitting a pull request, every single little step ends up in your repo.
I consider it a shortcoming that the creator of a pull request can no longer make changes—at least as long as there are no comments or a merge yet. After all, who wants to publicly document every single careless mistake?
Everything the AI ​​tells me about this is either complicated or doesn't work because I’m not the maintainer.
Please just squash the commits before you merge.

CDrummond:
Merged (and squashed) - thanks again!

PR #1235, opened 2026-06-19
Custom actions on streaming / app browse items

SimonArnold002:
# Custom actions on streaming / app browse items

## What & why
Custom actions currently only appear on **library** items. There is no way to add one to a **streaming/app** album or track while browsing a service (Qobuz/Tidal/Deezer/Bandcamp — new releases, an artist's discography, search), because that content is delivered as SlimBrowse `item_loop` items which bypass Material's custom-action wiring.

Two reasons online/app items get nothing today:
1. They bypass the `STD_ITEMS` action-menu path — their `stdItem` is `>= STD_ITEM_ONLINE_ARTIST` (300/301), so the `item.stdItem < STD_ITEMS.length` gate in `browse-functions.js` is false and they never reach the `CUSTOM_ACTIONS` handling.
2. The `item_loop` branch never sets `resp.itemCustomActions`.

Both are needed to render a custom action: the `CUSTOM_ACTIONS` marker in the item's own `i.menu`, **and** the view-level `resp.itemCustomActions`.

## Change
**`browse-resp.js`** (in the `item_loop` per-item handling): for online items and playable app rows, add the `CUSTOM_ACTIONS` marker to the item's `i.menu` and populate `resp.itemCustomActions`.

- Service **album rows** (e.g. Qobuz New Releases) carry no `favorites_url`/`metadata` — only a playable `item_id` and a two-line `title`/`subtitle` — so the rule keys off **playability** (`addedPlayAction && undefined==i.stdItem && !isFavorites && !isAppsTop`), and exposes `i.title`/`i.subtitle` as `i.album`/`i.artist` so `$ALBUMNAME`/`$ARTISTNAME` resolve.
- Category lookup tries `"<command>-<type>"` (the app's own view, e.g. `qobuz-album`) before the generic `"online-<type>"`. A category that is **defined at all** — even an empty array — wins, so an app/plugin can customise actions on its own view or suppress the generic ones (an empty category).

**`customactions.js`** (`doReplacements`): add `$SERVICE` (the browsing service, i.e. the `command`) and `$SUBTITLE` (the raw second line), and register them in `ACTION_KEYS`.

## Example `actions.json`
```json
{
  "online-album": [
    { "title": "Add to Listen to Later", "icon": "playlist_add",
      "lmscommand": ["listentolater","addctx","name:$ALBUMNAME","artist:$ARTISTNAME","svc:$SERVICE","image:$IMAGE"] }
  ],
  "listentolater-album": []
}
```
Browsing Qobuz → New Releases, an album's "…" now shows the action, passing the album name, the "Artist (Year)" subtitle, the service, and the cover. Inside the `listentolater` app's own list, the empty `listentolater-album` category overrides `online-album`, so the action isn't offered there.

## Notes
- Scoped to playable app/online rows; library items use the `*_loop` branches and are unaffected.
- With new `online-*` categories, zero behaviour change unless a user adds them to `actions.json`.
- Verified locally against `Release 6.4.3` on LMS 9.1 with the Qobuz plugin (album rows in browse lists, and a plugin's own app view).

### DIFF
--- a/MaterialSkin/HTML/material/html/js/browse-resp.js
+++ b/MaterialSkin/HTML/material/html/js/browse-resp.js
@@ -651,6 +651,43 @@ function parseBrowseResp(data, parent, options, cacheKey) {
                     i.menu.push(SHOW_IMAGE_ACTION);
                 }
 
+                // Surface custom actions on app/online (streaming) browse items. These arrive
+                // via item_loop (library content uses the *_loop branches) and bypass the
+                // STD_ITEMS action-menu path (online stdItems are >= STD_ITEM_ONLINE_ARTIST, so
+                // item.stdItem<STD_ITEMS.length is false), so add the CUSTOM_ACTIONS marker to
+                // the item's own menu and populate the view-level itemCustomActions. Service
+                // album rows (e.g. Qobuz New Releases) carry no favorites_url/metadata - only a
+                // playable item_id + title/subtitle - so key off playability and expose
+                // title/subtitle via item.album/item.artist so $ALBUMNAME/$ARTISTNAME resolve;
+                // $SERVICE carries the browsing service ('command').
+                let isOnlineAlbum  = STD_ITEM_ONLINE_ALBUM==i.stdItem;
+                let isOnlineArtist = STD_ITEM_ONLINE_ARTIST==i.stdItem;
+                let isAppItem = !isOnlineAlbum && !isOnlineArtist && !isOnlineTrack &&
+                                addedPlayAction && undefined==i.stdItem && !isFavorites && !isAppsTop;
+                if (isOnlineAlbum || isOnlineArtist || isOnlineTrack || isAppItem) {
+                    let btype    = isOnlineArtist ? "artist" : isOnlineTrack ? "track" : "album";
+                    let ocFilter = i.presetParams ? i.presetParams.favorites_url : undefined;
+                    // A plugin/app can define a custom-action category for its OWN view, named
+                    // "<command>-<type>" (e.g. "listentolater-album"). If defined - even as an
+                    // empty list - it takes precedence over the generic "online-*" category, so
+                    // a plugin's own list can show different actions, or none (an empty category
+                    // suppresses the generic actions on that app's items).
+                    let appCat = (undefined!=command) ? command+"-"+btype : undefined;
+                    let oca = (undefined!=appCat && undefined!=customActions && (appCat in customActions))
+                              ? getCustomActions(appCat, false, ocFilter)
+                              : getCustomActions("online-"+btype, false, ocFilter);
+                    if (undefined!=oca && oca.length>0) {
+                        if (isAppItem) {
+                            if (undefined==i.album)  { i.album   = i.title; }
+                            if (undefined==i.artist) { i.artist  = i.subtitle; }
+                            i.service = command;
+                        }
+                        addedDivider = addDivider(i, addedDivider);
+                        i.menu.push(CUSTOM_ACTIONS);
+                        if (undefined==resp.itemCustomActions) { resp.itemCustomActions = oca; }
+                    }
+                }
+
                 // Only show 'More' action if:
                 //    'more' is in baseActions and item has item_id
                 //    - OR -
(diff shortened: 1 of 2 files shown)


CDrummond:
Whilst I have no real objection to this, as I can see perhaps users might want to add their own actions here, it seems odd for a plugin to want this. The idea of these actions is that they are very much user specific.

Perhaps adding an extra (e.g.) "uiactions" field to the SlimBrowse response would be more portable? e.g.

1. Material sends "a" in its "features" tag - so this would be "features:ahi" (as Material would support these actions, headers, and infinite scroll).
2. When a plugin sees "a" in "features" it can add a "uiactions" to the response. e.g.

```json
"count":300,
"item_loop:"{ ... },
"uiactions":[
{"title":"Do some action", "icon":"playlisty_add", "command":["listentolater","addctx","name:$ALBUMNAME","artist:$ARTISTNAME","svc:$APP","image:$IMAGE"]}
]
```

...I'm using "$APP" here as its more generic than "$SERVICE"

Therefore user's don't need to update actions.json

SimonArnold002:
Thanks for the quick repsonse Craig, that's fair, and I agree `actions.json` is really the user's space; a plugin writing into it is intrusive. I'm happy to move plugin-owned actions to a `uiactions` field gated on `features:a`, and to use `$APP` rather than `$SERVICE`.

One thing worth flagging before I rework, because it affects whether `uiactions` can fully *replace* this PR or only complement it: `uiactions` in the response only lets a plugin decorate **its own** views. The motivating case here is "Add to Listen to Later" on a **Qobuz** album while browsing Qobuz — and that view is the Qobuz plugin's response. My plugin can't attach a `uiaction` to it: it doesn't own that response, and Qobuz has no knowledge of my plugin. The reason this PR reaches those rows at all is the global `online-*` categories — which is exactly the part you're (rightly) wary of.

So I think the two are complementary rather than either/or:

- **A plugin decorating its _own_ views** (e.g. my list's Remove/Move, or *not* offering "Add" inside it) → `uiactions` in the response. The plugin owns it and never touches `actions.json`.
- **Adding an action onto _another_ plugin's streaming view** (Add on Qobuz/Tidal while browsing) → the per-item online rendering in this PR, driven by a **user-defined** `actions.json` entry. That keeps it user-specific, which is your point.

If that split works for you, I'd rework as:

- **Material**: add the `uiactions` field + `features:a` + `$APP`, and **keep** the `browse-resp.js` per-item online rendering so a *user's* `actions.json` entry can target streaming rows.
- **My plugin**: stop writing `actions.json` entirely — use `uiactions` for its own view, and ship "Add on Qobuz" as a documented, opt-in `actions.json` snippet.

The one decision I'd defer to you on: whether to keep the per-item online rendering for user `actions.json` entries. Without it there's no way for a user to add an action to a streaming browse row at all (the original use case); but if you'd rather Material not support that, I'll scope this PR down to just `uiactions` and drop the `online-*` category rendering.

Happy to implement whichever direction you prefer.

CDrummond:
I can see your point. Perhaps I need to add a function to Material to allow plugins to add custom actions? Then these would be merged with a user's actions.json.

e.g. you'd call

```perl
MaterialSkin::Plugin::registerAction("online-album",  { "title": "Add to Listen to Later", "icon": "playlist_add",
      "lmscommand": ["listentolater","addctx","name:$ALBUMNAME","artist:$ARTISTNAME","svc:$SERVICE","image:$IMAGE"] });
```
...perl syntax will be wrong, as I'm not at my dev machine. But the idea being is that there'd be a function for you to call at start-up. Therefore no action required by users - and the syntax is the same as this PR.

I'll try and add the function sometime next week - as busy with non-coding things at the mo.

SimonArnold002:
Sounds good to me, and thanks. this woud also allow for Material to have  Add to Favourites to be available at album level for Streaming services like it is for local as currently its only available on track level for tracks not the album.

CDrummond:
The issue with "Add to favourites" is that it really needs to know if its already added so that its a "Remove from favourites" in that case, etc. But no way to do that at the mo.

SimonArnold002:
Okay. Using Svens fork of Qobuz he managed to ad an  Add to Favourites option  at album level. When in Favourites view you could remove them like local favourites.

PR #1241, opened 2026-07-10
Don't pass artist_id when browsing album from queue and np

darrell-k:
I think the album link from the queue/now playing/MAI screen should show the whole album, not filter by the current (track) artist.

### DIFF
--- a/MaterialSkin/HTML/material/html/js/itemlinks.js
+++ b/MaterialSkin/HTML/material/html/js/itemlinks.js
@@ -19,10 +19,9 @@ function show_artist(event, id, title, page) {
     browseItem(event, ["albums"], ["artist_id:"+id, ARTIST_ALBUM_TAGS, SORT_KEY+ARTIST_ALBUM_SORT_PLACEHOLDER], unescape(title), page);
 }
 
-function showAlbum(event, album_id, title, page, subtitle) {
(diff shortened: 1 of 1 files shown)


CDrummond:
Not sure. Would the following not be better??

```javascript
function showAlbum(event, album_id, artist_id, title, page, subtitle) {
    let cmd = ["album_id:"+album_id, trackTags(true), SORT_KEY+"tracknum"];
    if (lmsOptions.noArtistFilter) {
        cmd.params.push('material_skin_artist_id:'+artist_id);
    } else {
        cmd.params.push('artist_id:'+artist_id);
    }

    browseItem(event, ["tracks"], cmd, unescape(title), page, undefined==subtitle ? subtitle : unescape(subtitle));
}
```

darrell-k:
I thought about that, but my view is that when clicking on an album title in the queue, you're not explicitly selecting an artist. So it doesn't really fit with the artist filter: "When browsing an artist's albums use all tracks from compilation albums, albums this artist appears on, etc., or only show tracks from the current artist."

Also, when using the "toggled" queue with album headers, where the link refers to multiple tracks, it would only use the artist from the first track.

CDrummond:
Fair point, then perhaps:

```javascript
function showAlbum(event, album_id, artist_id, title, page, subtitle) {
    browseItem(event, ["tracks"], ["album_id:"+album_id, "material_skin_artist_id:"+artist_id, trackTags(true), SORT_KEY+"tracknum"], unescape(title), page, undefined==subtitle ? subtitle : unescape(subtitle));
}
```

PR #1245, opened 2026-07-27
Switch from JSON::XS::VersionOneAndTwo to JSON::XS

mavit:
LMS no-longer uses `JSON::XS::VersionOneAndTwo` internally, and it’s simpler to use `JSON::XS` directly.


CDrummond:
Thanks!

PR #1255, opened 2026-08-24
Fix image url's rewrite when using https proxy

andrew-codechimp:
As reported within AF's virtual library creator plugin
https://github.com/AF-1/lms-virtuallibrarycreator/issues/15

Images were not correctly displayed due to the image path rewrite only looking for http, not https.
Changed startswith to look for either.

Tested on my own setup and seems not to have any detrimental effect but I'm not a JS developer.


CDrummond:
Wow! Thanks for looking at this - must admit it totally slipped my mind!

andrew-codechimp:
No problem, I was bored and wanted a challenge.

PR #1261, opened 2026-09-05
Show command result text for now playing alt buttons

alcoheca:
repeatClicked/shuffleClicked pass the button's static pre-click tooltip to doCommand as the post-command message, so the toast always showed what the button would do, never what it just did. Now prefers the command response's item_loop[0].text when present.

This matches behaviour on Squeezeplay


CDrummond:
Thanks!

PR #1262, opened 2026-09-05
Add outline/disabled jiveStyle icon mappings

alcoheca:
repeat/shuffle alt buttons only mapped thumbsUp/thumbsDown/love to a Material Icons glyph, not their off-state counterparts (hate, thumbsUpDisabled, thumbsDownDisabled), which fell back entirely to the plugin's own icon image. Adds favorite_border/thumb_up_off_alt/ thumb_down_off_alt, already present in the bundled font.


CDrummond:
Thanks again!

PR #1268, opened 2026-09-21
Song Info - Render string/text new line as html break line

cucko:
Render string/text new line as html break line - helpful when comment is enabled to be shown in song info, if comment contains new lines.
current
Screenshot 2026-09-21 182747
updated css
Screenshot 2026-09-21 182734


CDrummond:
Thanks!

